PR #3, opened 2022-06-22
Added build.bat script to build cirrus.exe

lukehb:
Also made some changes to server of the "home page" so that those paths are found correctly when using the compiled cirrus.exe


mentalnote (approved):
lgtm

PR #5, opened 2022-06-30
Add ability to limit the number of connections (default: unlimited)

Belchy06:
Some customers have been requesting the ability to limit the number of concurrent sessions, modify the config.json to set the limit for the number of sessions (-1 = unlimited connections)


lukehb (approved):
lgtm

PR #8, opened 2022-07-08
Fix controller button messages not being sent

sheepsteak:
At the moment pressing and releasing gamepad buttons doesn't send any data back to UE. With this fix the messages are sent.

However, the triggers on an Xbox gamepad don't seem to work still. The messages send (and I can log them out in UE) but they don't have the same result as connecting a gamepad to UE locally.

### DIFF
--- a/SignallingWebServer/scripts/app.js
+++ b/SignallingWebServer/scripts/app.js
@@ -168,6 +168,7 @@ function emitControllerButtonPressed(controllerIndex, buttonIndex, isRepeat) {
     Data.setUint8(1, controllerIndex);
     Data.setUint8(2, buttonIndex);
     Data.setUint8(3, isRepeat);
+    sendInputData(Data.buffer);
 }
 
 function emitControllerButtonReleased(controllerIndex, buttonIndex) {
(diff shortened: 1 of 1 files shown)


lukehb (approved):
Thanks! For the other issue about triggers not working as expected this sounds like a UE side issue. We will open a UE jira for this internally. @Belchy06 For the UE side issue.

Belchy06:
Hey there @sheepsteak.

I've been looking into the shoulder issue you brought up in your PR.

I'm unable to reproduce this issue using the latest version of Pixel Streaming found in UE5-Main (https://github.com/EpicGames/UnrealEngine/tree/ue5-main).

I'm not sure what version you were using, and with a refactor recently being completed on how Pixel Streaming handles input from the browser it's likely this issue was fixed then.

Let me know how you go!

sheepsteak:
Thanks, @Belchy06. They seem to not work in 5.0.3 and earlier. It's like it was mapped to the wrong thing (possibly right shoulder button?).

We tried `ue5-main` ourselves just now and it looks like it's working as expected 👍🏻

PR #15, opened 2022-07-12
Add a deadzone to the analog axes

Belchy06:
Now that controller inputs are working again, it has been noted that the front end will constantly emit small analog inputs (0.00xx when the scale is -1 -> 1). 

Adding a deadzone to these inputs allows for more comfortable use of a controller with PixelStreaming.


Belchy06:
Closing this as the engine already has configuration options for dead zones.

image

PR #17, opened 2022-07-13
Added support for handling websocket messages sent as binary.

lukehb:
Websockets in browsers and elsewhere support sending websocket messages a text and binary. Text support has been in Pixel Streaming javascript since initial release; however binary support has not been.

This commit adds support for parsing binary messages, which may be useful for alternate signalling server implementations.


mentalnote (commented):
lgtm

mentalnote (approved):

PR #18, opened 2022-07-17
Fix syntax errors

sheepsteak:
Just a few fixes for some syntax errors that start to cause problems when bundling the browser-side code.


mentalnote (approved):
lgtm

PR #19, opened 2022-07-18
Remove hardcoded protocol

Belchy06:
Currently, if a user were to add a new message type to pixel streaming, they need to update both the front end and the c++ to support this new message type.

This PR adds the ability for users to simply register their new message in the c++ and all peers connected will simply receive the updated protocol and keep working as intended.

We keep a populated default protocol to maintain backwards compatibility with previous engine versions.


SignallingWebServer/scripts/app.js
```diff
@@ -108,99 +104,57 @@ function updateStatus() {
             // Button 6 is actually the left trigger, send it to UE as an analog axis
             // Button 7 is actually the right trigger, send it to UE as an analog axis
             // The rest are normal buttons. Treat as such
-            if (currButton.pressed && !prevButton.pressed) {
```
mentalnote, line 110: what's the purpose of this change?

SignallingWebServer/scripts/app.js
```diff
@@ -108,99 +104,57 @@ function updateStatus() {
...
+            if (currButton.pressed) {
                 // New press
                 if (i == 6) {
```
mentalnote: Should really replace all these magic numbers with variables that represent what they are

SignallingWebServer/scripts/app.js
```diff
@@ -1485,59 +1419,247 @@ function onOrientationChange(event) {
...
+        "structure": [ "uint8", "uint8", "uint8" ]
+    },
+    "GamepadButtonAnalog": {
+        "id": 92,
+        "byteLength": 10,
+        //            ctrlerId   button  analogValue
+        "structure": [ "uint8", "uint8", "double" ]
+    },
+
+    // Old EToPlayerMsg enum
```
mentalnote: What's the difference between these key-value entries and the ones where you additionally set `byteLength: 0` and `structure: []`?

SignallingWebServer/scripts/app.js
```diff
@@ -2389,7 +2454,8 @@ function clearMouseEvents(playerElement) {
         playerElement.onmousewheel = null
     } else {
         playerElement.removeEventListener('DOMMouseScroll', function(e) {
-            emitMouseWheel(e.detail * -120, e.offsetX, e.offsetY);
+            let coord = normalizeAndQuantizeUnsigned(e.offsetX, e.offsetY);
+            sendInputMessage("MouseWheel", [ e.detail * -120, coord.x, coord.y ]);
```
mentalnote: Should give this -120 a variable and name that represents its purpose

SignallingWebServer/scripts/app.js
```diff
@@ -2038,19 +2113,8 @@ function registerTouchEvents(playerElement) {
...
-            byte += 2;
-            data.setUint16(byte, coord.y, true);
-            byte += 2;
-            data.setUint8(byte, fingerIds[touch.identifier], true);
-            byte += 1;
-            data.setUint8(byte, 255 * touch.force, true); // force is between 0.0 and 1.0 so quantize into byte.
-            byte += 1;
-            data.setUint8(byte, coord.inRange ? 1 : 0, true); // mark the touch as in the player or not
-            byte += 1;
+            sendInputMessage(type, [ 1, coord.x, coord.y, fingerIds[touch.identifier], 255 * touch.force, coord.inRange ? 1 : 0 ]);
```
mentalnote: Give the `1` at the start of the created array here a variable and name describing its purpose along with the `255` (MaxByteValue)

mentalnote: Actually, shouldn't that `1` be `touches.length`?

mentalnote (dismissed):

SignallingWebServer/scripts/app.js
Belchy06: Just code cleanliness. We had a lot of if/else to check if the this event should be emitted as a repeat press or not.

I found the only difference between this first if block and the last elseif was the value of the last parameter to emitControllerButtonPressed (L118 / L136)

SignallingWebServer/scripts/app.js
Belchy06: With the rewrite of the UE side protocol, we are unable to support message like this where there may be a varying amount of information unless it's parsed as a string (eg the descriptors). To maintain backwards compatibility we now send each touch individually and such the touches array now only has a length of 1

SignallingWebServer/scripts/app.js
Belchy06: This has been fixed by instead using the standardized `onwheel` event instead of `onmousewheel` for chrome and `DOMMouseScroll` for firefox.

SignallingWebServer/scripts/app.js
Belchy06: the ones below are for receiving messages from UE in which case they are always formatted as a Uint8Array. All the below messages are handled in `webRtcPlayerObj.onDataChannelMessage`

SignallingWebServer/scripts/app.js
```diff
@@ -1485,59 +1447,247 @@ function onOrientationChange(event) {
...
+        "id": 61,
+        "byteLength": 1,
+        //             keyCode
+        "structure": [ "uint8" ]
+    },
+    "KeyPress": {
+        "id": 62,
+        "byteLength": 2,
+        //             charcode
+        "structure": [ "uint16" ]
```
lukehb: Seems a little odd that KeyDown/Up has different size to KeyPress

SignallingWebServer/scripts/app.js
```diff
@@ -2027,31 +2128,17 @@ function registerTouchEvents(playerElement) {
...
-            byte += 2;
-            data.setUint16(byte, coord.y, true);
-            byte += 2;
-            data.setUint8(byte, fingerIds[touch.identifier], true);
-            byte += 1;
-            data.setUint8(byte, 255 * touch.force, true); // force is between 0.0 and 1.0 so quantize into byte.
-            byte += 1;
-            data.setUint8(byte, coord.inRange ? 1 : 0, true); // mark the touch as in the player or not
-            byte += 1;
+            sendInputMessage(type, [ numTouches, coord.x, coord.y, fingerIds[touch.identifier], MaxByteValue * touch.force, coord.inRange ? 1 : 0 ]);
```
lukehb: Have you verified sending touches in this way preserves multi-touch capabilities, like pinch/zoom?

SignallingWebServer/scripts/app.js
```diff
@@ -949,8 +906,14 @@ function setupWebRtcPlayer(htmlElement, config) {
 
     webRtcPlayerObj.onDataChannelMessage = function(data) {
```
lukehb, line 1342: I was hoping here we could end up with some API surface in the frontend like so:

`RegisterInputProtocolHandler("FreezeFrame", OnFreezeFrameMessage);`

Then when we get the protocol message we handle it by having already registered the following:

`RegisterInputProtocolHandler("Protocol", OnProtocolMessage);`

where OnProtocolMessage looks very roughly like so:

```
function onProtocolMessage(protocolJSON){
    this.protocolJSON = protocolJSON
    this.protocolHandlers = {}
    for(let protocol of protocolJSON){
        if(protocol in this.registeredProtocolHandlers){
            this.protocolHandlers[protocol] = this.registeredProtocolHandlers[protocol];
        }
        else {
            console.error(`There was no registered protocol handler for ${protocol} - try adding one using RegisterInputProtocolHandler(${protocol}, myHandler)`);
        }
    }
}
```

And ofc for backwards compatibility we will pre-populate the protocol handlers with all the defaults we have supported in the past versions.

SignallingWebServer/scripts/app.js
lukehb: Should the protocol differentiate between streamer and player message types? ToStreamer/FromStreamer - it seems like it could be helpful? If they are split out like so I think it becomes more kosher to have these ones below that have no structure/byte length. Is it worth saying structure is uint8 array for consistency?

SignallingWebServer/scripts/app.js
```diff
@@ -1485,59 +1447,247 @@ function onOrientationChange(event) {
     }, 500);
 }
 
-// Must be kept in sync with PixelStreamingProtocol::EToUEMsg C++ enum.
-const MessageType = {
+function sendInputMessage(messageType, indata = []) {
```
lukehb: I would consider calling this sendInputProtocolMessage()

SignallingWebServer/scripts/app.js
```diff
@@ -1485,59 +1447,247 @@ function onOrientationChange(event) {
...
 
+    data.setUint8(0, messageFormat.id);
+    byteOffset = 1;
+
+    indata.forEach((element, idx) => {
+        type = messageFormat.structure[idx];
+        switch(type) {
+            case "uint8":
+                data.setUint8(byteOffset, element);
+			    byteOffset += 1;
```
lukehb: weird whitespace or just me?

SignallingWebServer/scripts/app.js
```diff
@@ -1485,59 +1447,247 @@ function onOrientationChange(event) {
...
-    FpsRequest: 2,
-    AverageBitrateRequest: 3,
-    StartStreaming: 4,
-    StopStreaming: 5,
-    LatencyTest: 6,
-    RequestInitialSettings: 7,
-
-    /**********************************************************************/
-
+    "IFrameRequest": {
```
lukehb: Does IFrameRequest work atm? If not, can we make it work?

If it does work, let's hook up a button for this in the settings panel, this seems to come up often enough where people want to request a keyframe from browser side.

SignallingWebServer/scripts/app.js
```diff
@@ -346,8 +328,8 @@ function setupHtmlEvents() {
             let minQP = document.getElementById('encoder-min-qp-text').value;
             let maxQP = document.getElementById('encoder-max-qp-text').value;
 
-            emitCommand({ "Encoder.MinQP": minQP });
-            emitCommand({ "Encoder.MaxQP": maxQP });
+            emitDescriptor("ToStreamerCommand", { "Encoder.MinQP": minQP });
```
lukehb: Instead of hardcoded strings, these should probably come from a DefaultProtocolKeys.ToStreamerCommand or something like that.

lukehb (dismissed):
Left some comments mostly about changing how the API surface looks.

SignallingWebServer/scripts/app.js
```diff
@@ -65,25 +63,455 @@ let afk = {
...
+    registerMessageHandler(MessageDirection.ToStreamer, "MouseDown", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "MouseUp", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "MouseMove", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "MouseWheel", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "TouchStart", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "TouchEnd", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "TouchMove", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "GamepadButtonPressed", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "GamepadButtonReleased", sendMessageToStreamer);
+    registerMessageHandler(MessageDirection.ToStreamer, "GamepadAnalog", sendMessageToStreamer);
```
lukehb, line 340: Beautiful

SignallingWebServer/scripts/app.js
```diff
@@ -65,25 +63,455 @@ let afk = {
...
+function onInputControlOwnership(data) {
+    let view = new Uint8Array(data);
+    let ownership = view[1] === 0 ? false : true;
+    console.log("Received input controller message - will your input control the stream: " + ownership);
+    inputController = ownership;
+}
+
+function onProtocolMessage(data) {
+    /**
+     * TODO
```
lukehb: Why todo?

SignallingWebServer/scripts/app.js
```diff
@@ -65,25 +63,455 @@ let afk = {
...
+        }
+        let direction = protocolJSON.direction;
+        delete protocolJSON.direction;
+        Object.keys(protocolJSON).forEach((messageType) => {
+            let message = protocolJSON[messageType];
+            switch (direction) {
+                case MessageDirection.ToStreamer:
+                    // Check that the message contains all the relevant params
+                    if (!message.hasOwnProperty("id") || !message.hasOwnProperty("byteLength")) {
+                        console.error(`ToStreamer->${messageType} protcol definition was malformed`);
```
lukehb: Error message should mention it was malformed because it didn't contain both id and byteLength and should print what the message was.

lukehb: Same with structure parsing below

SignallingWebServer/scripts/app.js
```diff
@@ -65,25 +63,455 @@ let afk = {
...
+                        // If we've registered a handler for this message type we can add it to our supported messages. ie registerMessageHandler(...)
+                        toStreamerMessages.add(messageType, message);
+                    } else {
+                        console.error(`There was no registered handler for "${messageType}" - try adding one using registerMessageHandler(MessageDirection.ToStreamer, "${messageType}", myHandler)`);
+                    }
+                    break;
+                case MessageDirection.FromStreamer:
+                    // Check that the message contains all the relevant params
+                    if (!message.hasOwnProperty("id")) {
+                        console.error(`FromStreamer->${messageType} protcol definition was malformed`);
```
lukehb: Same as above comment, also typo in protocol

SignallingWebServer/scripts/twoWapMap.js
```diff
@@ -0,0 +1,23 @@
...
+    add(key, value) {
+        this.map[key] = value;
+        this.reverseMap[value] = key;
+    }
+
+    remove(key, value) {
+        delete this.map[key];
+        delete this.reverseMap[value];
+    }
+}
```
lukehb: Let's keep this in app.js until we introduce a bundler

SignallingWebServer/scripts/app.js
Belchy06: Verified that multi-touch still works with UEs inbuilt on-screen joysticks

SignallingWebServer/scripts/app.js
Belchy06: It does now!

lukehb (approved):
lgtm

PR #20, opened 2022-07-26
Adding offer to receive support.

mcottontensor:
Added a flag to app.js to allow the player to offer to receive.
Added support in cirrus to handle players sending offers to receive.


lukehb (approved):
lgtm

PR #23, opened 2022-08-08
Cirrus logging UE version

mentalnote:
When the signalling server logs what version of UE it is for it now correctly says 4.27.


mentalnote:
Something went wrong here...

PR #24, opened 2022-08-08
Update cirrus logging to refer to UE version 4.27

mentalnote:
When the signalling server logs what version of UE it is for it now correctly says 4.27


lukehb (approved):
lgtm

PR #25, opened 2022-08-08
UE4.26 compatibility fixes

mentalnote:
All changes necessary for PixelStreaming Infrastructure to work with UE4.26 PixelStreaming (uses UE4.27 branch changes as a base).

Note that Linux scripts have been removed as UE4.26 does not support PixelStreaming on Linux.


lukehb (approved):
lgtm

PR #27, opened 2022-08-11
Add handling for mouse double click

Belchy06:
Currently, an Unreal application (or even editor) is unable to handle double clicks and instead registers them as two individual clicks.

This PR fixes that, but also requires the use of the latest version of the Pixel Streaming plugin.


mentalnote (approved):
lgtm

PR #33, opened 2022-09-30
Fix afk overlay on mobile devices

Belchy06:


lukehb (approved):
lgtm - this should fix on devices where there is no PointerLock, e.g. mobile

PR #39, opened 2022-10-14
Replaced hardcoded log path with given parameter path

Mirmidion:
Specifying a path with registering the file logger only created the folder, but never used it in creating the write stream.
Fixed it so that the parameter gets used in creating the write stream.


lukehb (approved):
lgtm

lukehb:
Your PR will need to commit be code signed. See here: https://docs.github.com/articles/about-gpg/

Mirmidion:
Signed the commit

Issue #46, opened 2022-11-18
KeyboardEvent.keyCode is Deprecated


xosg:
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

`Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.`

lukehb:
CC @Belchy06 - could you take a look at this one?

PR #48, opened 2022-11-24
Added check if KeyboardEvent.keyCode deprecated then use KeyboardEvent.code + mapping instead.

lukehb:
I "fixed" this by adding a check to see if KeyboardEvent.keyCode is not present, if so, I use KeyboardEvent.code. This change ends up sending the exact same keyCodes as before because I created a mapping from KeyboardEvent.code to KeyboardEvent.keyCode. This will mean the deprecated method will continue to be used and the codepath will be exactly the same as it has been until such a time when the browsers actually remove KeyboardEvent.keyCode and then this new code will execute. 

This means no change at all on the UE side!

### DIFF
--- a/SignallingWebServer/scripts/app.js
+++ b/SignallingWebServer/scripts/app.js
@@ -2442,9 +2442,134 @@ const SpecialKeyCodes = {
     RightAlt: 255
 };
 
-// We want to be able to differentiate between left and right versions of some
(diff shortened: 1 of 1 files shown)


lukehb:
This PR addresses #46

Belchy06 (approved):
LGTM!

PR #53, opened 2022-12-14
Input re-structuring

Belchy06:
This PR moves to a transceiver model to allow for easier bi-directional communication.
 
Instead of having explicit ToStreamer and FromStreamer protocols, we now have a unified protocol with each message type capable of supporting bi-directional messaging. 

NOTE: This is a breaking change from 5.1.x and will require users using a version of UE >5.1.x to use the new input system.


mentalnote (changes requested):
If the whitespace/formatting changes are intentional, could they be split out into their own commit?

Belchy06:
Closed as it's no longer needed

Issue #50, opened 2022-11-29
Wrong path to login page when using authentication


RadAlex:
in cirrus.js, when using authentication, the path is:
```
if(config.UseAuthentication){
	if(config.EnableWebserver) {
		app.get('/login', function(req, res){
			res.sendFile(__dirname + '/login.htm');
		});
	}
```

But the login page is _login.htm**l**_
And it seems not to be able to find the `public` folder, so address should look like this for it to work:
```
res.sendFile(__dirname + '/public/login.html');
```

PR #55, opened 2022-12-18
Fix incorrect login page path when using authentication

Belchy06:
Thanks to @RadAlex for mentioning this in https://github.com/EpicGames/PixelStreamingInfrastructure/issues/50


mentalnote (approved):
Looks good

PR #57, opened 2023-01-01
[Bug Fixed] Insufficient validation when decoding a Socket packet

iot-defcon:
Signed-off-by: Sergev ₱ <118327710+iot-defcon@users.noreply.github.com>

### Description The Summary
Due to improper type validation in the `socket.io-parser` library (which is used by the socket.io and `socket.io-client` project to encode and decode Socket.IO packets), it is possible to overwrite the _placeholder object which allows an attacker to place references to functions at arbitrary places in the resulting query object.

**PoCs By DefC0n**
```js
const decoder = new Decoder();

decoder.on("decoded", (packet) => {
 console.log(packet.data); // prints [ 'hello', [Function: splice] ]
})

decoder.add('51-["hello",{"_placeholder":true,"num":"splice"}]');
decoder.add(Buffer.from("world"));
```
This bubbles up in the socket package:
```js
io.on("connection", (socket) => {
 socket.on("pwned", (val) => {
 // here, "val" could be a function instead of a buffer
 });
});
```
You need to make sure that the payload that you received from the client is actually a Buffer object:
```js
io.on("connection", (socket) => {
 socket.on("PWNED", (val) => {
 if (!Buffer.isBuffer(val)) {
 socket.disconnect();
 return;
 }
 // 
 });
});
```
If that's already the case, then you are not impacted by this issue, and there is no way an attacker could make your server crash (or escalate privileges, ...).

vulnerable of values that could be sent by a malicious user:
 * a number that is out of bounds

vulnerable packet: `451-["pwned",{"_placeholder":true,"num":10}]`
```js
io.on("connection", (socket) => {
 socket.on("hello", (val) => {
 // val is `undefined`
 });
});
```

CVE-2022-2421
[CWE-20](https://cwe.mitre.org/data/definitions/20.html)
[CWE-1287](https://cwe.mitre.org/data/definitions/1287.html)
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`

### DIFF
--- a/Matchmaker/package-lock.json
+++ b/Matchmaker/package-lock.json
@@ -696,9 +696,9 @@
       "integrity": "sha512-Qd/iwn3VskrpNO60BeRyCyr8ZWw9CPZyitW4AQwmRZ8zCiyDiL+znRnWX6tDHXnWn1sJrM1+b6Mn6wEDJJ4aYQ=="
     },
     "node_modules/socket.io-parser": {
-      "version": "4.0.4",
-      "resolved": "https://registry.npmjs.org/socket.io-parser/-/socket.io-parser-4.0.4.tgz",
-      "integrity": "sha512-t+b0SS+IxG7Rxzda2EVvyBZbvFPBCjJoyHuE0P//7OAsN23GItzDRdWa6ALxZI/8R5ygK7jAR6t028/z+7295g==",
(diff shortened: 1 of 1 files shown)


mentalnote (approved):
verified, looks good, merging.

PR #60, opened 2023-01-10
Handle "defaultToHover" field in offer

StomyPX:
This pull request allows the infrastructure to handle an extra field in the initial connection offer for determining whether the frontend should default to hovering mouse instead of captured.

Intended for use with editor streaming, this field has been put in the connection offer instead of its own config message to avoid an error on older infrastructure versions which will not be expecting new message types.

### DIFF
--- a/SignallingWebServer/scripts/app.js
+++ b/SignallingWebServer/scripts/app.js
@@ -61,6 +61,8 @@ let responseEventListeners = new Map();
 let freezeFrameOverlay = null;
 let shouldShowPlayOverlay = true;
 
+let defaultToHover = false;
+
 let isFullscreen = false;
 let isMuted = false;
 // A freeze frame is a still JPEG image shown instead of the video.
@@ -697,7 +699,7 @@ function onFullscreenChange() {
 
 function parseURLParams() {
     let urlParams = new URLSearchParams(window.location.search);
-    inputOptions.controlScheme = (urlParams.has('hoveringMouse') ?  ControlSchemeType.HoveringMouse : ControlSchemeType.LockedMouse);
+    inputOptions.controlScheme = (!!urlParams.get('hoveringMouse') ?  ControlSchemeType.HoveringMouse : ControlSchemeType.LockedMouse);
     let schemeToggle = document.getElementById("control-scheme-text");
     switch (inputOptions.controlScheme) {
         case ControlSchemeType.HoveringMouse:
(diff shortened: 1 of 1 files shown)


SignallingWebServer/scripts/app.js
```diff
@@ -2733,6 +2742,11 @@ function connect() {
             if (!UrlParamsCheck('offerToReceive')) {
                 onWebRtcOffer(msg);
             }
+            defaultToHover = !!msg.defaultToHover;
```
lukehb, line 2745: I know its much more verbose, but for code consistency reasons can you change any !! to

Object.prototype.hasOwnProperty.call(foo, 'bar');

SignallingWebServer/scripts/app.js
```diff
@@ -697,7 +699,7 @@ function onFullscreenChange() {
 
 function parseURLParams() {
     let urlParams = new URLSearchParams(window.location.search);
-    inputOptions.controlScheme = (urlParams.has('hoveringMouse') ?  ControlSchemeType.HoveringMouse : ControlSchemeType.LockedMouse);
+    inputOptions.controlScheme = (!!urlParams.get('hoveringMouse') ?  ControlSchemeType.HoveringMouse : ControlSchemeType.LockedMouse);
```
lukehb, line 702: I know its much more verbose, but for code consistency reasons can you change any !! to

Object.prototype.hasOwnProperty.call(foo, 'bar');

SignallingWebServer/scripts/app.js
StomyPX: For this one specifically, it needs to also be false if it has the property exists but is in fact false. So it will need to be something like (Assuming I wrote it right, I rarely use JS and I know it has some weird comparison syntax compared to C/C++)
```js
if (msg.hasOwnProperty("defaultToHover") && msg.defaultToHover == true) {
     defaultToHover = true;
} else {
     defaultToHover = false;
}
```
Is that okay?

SignallingWebServer/scripts/app.js
StomyPX: Actually it might have been this one that needs to check for exists and is true, I think the other will work just checking for existence. Though it does pain me to change it to that as it's like using `#ifdef` for when you actually mean `#if`.

StomyPX: Alternatively I could just use Object.is() for both if that's alright

lukehb: Fair point!

lukehb: Landing as is

lukehb (approved):
Have reviewed. Approved.

PR #63, opened 2023-01-18
Convert frontend javascript to typescript and refactor.

lukehb:
This PR addresses. #9.

---

The reference frontend code that ships in this repository is intended to act as a starting point that developers can modify and extend to suit the needs of their Pixel Streaming applications. Unfortunately, the structure of the existing JavaScript code does not enforce a clean separation between the base code and any custom modifications, intermingling everything in a way that makes it difficult to change either the base code or the modifications in isolation. This makes the process of porting customisations from one version of the frontend to another a tedious and labour-intensive process, significantly hindering developers' ability to update their code when new versions of the frontend are released. This is particularly problematic when considering the fact that new releases of web browsers can break the frontend (and this has happened multiple times in the past), so the ability to consume updates as quickly and easily as possible is critical to ensuring that Pixel Streaming applications remain accessible to all users when deployed in a production environment.

To address these limitations, this pull request draws on code from the [TypeScript frontend implementation](https://github.com/ScalablePixelStreaming/Frontend) used by [Scalable Pixel Streaming](https://scalablestreaming.io/) and merges it into the upstream reference implementation, with additional changes and improvements to ensure parity with the functionality present in the existing JavaScript implementation. The Scalable Pixel Streaming frontend code was contributed by my colleagues [Adam Rehn](https://github.com/adamrehn), [Adrian Zahra](https://github.com/adrianZahra), [Daniel Holden](https://github.com/dan-tw) and [David MacPherson](https://github.com/david-macpherson) from TensorWorks.

The new TypeScript library is a modular and extensible package that has been explicitly designed to ensure a clean separation between the base code and any custom modifications. The new TypeScript implementation acts as a drop-in replacement for the existing JavaScript implementation and introduces the following benefits:

- **Works with modern web development workflows:** once the accompanying GitHub Actions CI/CD workflows are in place, the TypeScript library will be provided as both an [NPM](https://www.npmjs.com/) package through [GitHub Packages](https://github.com/orgs/EpicGames/packages) and a [UMD](https://github.com/umdjs/umd) module through [this repository's releases](https://github.com/EpicGames/PixelStreamingInfrastructure/releases), making it easy to consume the library from either TypeScript code or plain JavaScript code using modern web development tools and workflows.

- **Easy to integrate with frontend web frameworks:** all of the core logic in the library can be overridden by extending the base `Application` class and providing alternate implementations for the relevant methods. This makes it straightforward to write derived implementations that integrate the library with anything from plain [DOM](https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model) APIs all the way through to complex frontend web frameworks. The library ships with a default implementation that demonstrates the use of the base `Application` class with its default behaviour, which can be used as a drop-in replacement for the existing HTML, CSS and JavaScript shipped in this repository.

- **Easy to update:** the clean separation between the core logic and custom derived implementations allows developers to rapidly update to new releases of the frontend and stay on top of any WebRTC changes in new web browser updates. Simply update the dependency version where you reference the library (e.g. build configuration files or a `<script>` tag) and you're good to go.

- **Supports custom protocol extensions:** the TypeScript library makes it possible to extend the WebSocket protocol used to communicate with the Pixel Streaming signalling server, by providing hooks for registering custom message types. This allows downstream projects that use custom signalling servers (e.g. Scalable Pixel Streaming) to consume the reference frontend library whilst still retaining support for their specific protocol extensions.


adamrehn (approved):

Frontend/implementations/EpicGames/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "module": "es6",
+    "esModuleInterop": true,
+    "target": "es5",
+    "moduleResolution": "node",
+    "sourceMap": false,
+    "allowJs": true,
+    "declaration": false
+  },
+  "lib": ["es2015"],
+  "include": ["./src/*.ts", "index.d.ts", "../../library/src/Ui/LatencyTest.ts", "../../library/src/Ui/VideoQpIndicator.ts", "../../library/src/Ui/OnScreenKeyboard.ts", "../../library/src/Overlay/Overlays.ts", "../../library/src/Ui/Controls.ts"],
```
hmuurine: Could use tsconfig project references to avoid listing the library ts files. You can refer to another directory that has a `tsconfig.json` file and it will get built according to the referred config file rules.
```suggestion
  "include": ["./src/*.ts", "index.d.ts"],
  "references": [
    { "path": "../../library" }
  ],
```

See: https://www.typescriptlang.org/docs/handbook/project-references.html

Frontend/library/src/AFK/AFKController.ts
```diff
@@ -0,0 +1,130 @@
...
+        // instantiate a new overlay 
+        this.showAfkOverlay();
+
+        // update our countDown timer and overlay contents
+        this.countDown = this.closeTimeout;
+        this.countdownActive = true;
+        this.afkOverlay.updateCountdown(this.countDown)
+
+        // if we are in locked mouse exit pointerlock
+		if (!this.config.isFlagEnabled(Flags.HoveringMouseMode)) {
```
hmuurine:
```suggestion
        if (!this.config.isFlagEnabled(Flags.HoveringMouseMode)) {
```

Frontend/library/src/AFK/AFKController.ts
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }, 1000);
+    }
+
+    /**
+     * An override method for showing the afk overlay 
+     */
+    showAfkOverlay() { 
+		// Base Functionality: Do Nothing
+	}
```
hmuurine:
```suggestion
    }
```

Frontend/library/src/AFK/AFKController.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * An override method for hiding the afk overlay 
+     */
+    hideCurrentOverlay() { 
+		// Base Functionality: Do Nothing
+	}
```
hmuurine:
```suggestion
    }
```

Frontend/library/src/AFK/AFKOverlay.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright Epic Games, Inc. All Rights Reserved.
+
+import { ActionOverlay } from "../Overlay/ActionOverlay";
+
+/**
+ * Show an overlay for when the session is unattended, it begins a countdown timer, which when elapsed will disconnect the stream.
+ */
+export class AFKOverlay extends ActionOverlay {
+	/**
```
hmuurine: [minor] Looks like this file is using tabs for indentation whereas some other files use 4 spaces. And in some files some rows have spaces and others tabs. Perhaps could unify the files to use either tabs or spaces, and not mix styles?

Even better, maybe could run the files through Prettier or some other auto-formatter for consistency?

Frontend/library/src/Inputs/KeyboardController.ts
```diff
@@ -0,0 +1,254 @@
...
+        // Backspace is not considered a keypress in JavaScript but we need it
+        // to be so characters may be deleted in a UE text entry field.
+        if (keyCode === SpecialKeyCodes.backSpace) {
+            document.dispatchEvent(new KeyboardEvent('keypress', { charCode: SpecialKeyCodes.backSpace }));
+        }
+
+		if (this.config.isFlagEnabled(Flags.SuppressBrowserKeys) && this.isKeyCodeBrowserKey(keyCode)) {
+            keyboardEvent.preventDefault();
+        }
+    }
```
hmuurine: Would it make sense to have a config item for temporarily suspending sending events to the pixel streaming app? Consider the use case where the pixel streaming app is embedded as a part of hybrid web application, and the user opens some kind of web config overlay / modal or something similar. In that case we might want to not send events to the pixel streaming app until the modal is closed.

Of course this could be achieved by a bunch of event handlers attached to some kind of full screen modal backdrop that prevents the events propagating down to the document level, so can be worked around. But maybe could consider a config option that could be modified on the fly to make it easier for the devs?

Frontend/library/src/VideoPlayer/VideoPlayer.ts
```diff
@@ -0,0 +1,198 @@
...
+
+        this.onResizePlayerCallback = () => { console.log("Resolution changed, restyling player, did you forget to override this function?"); }
+        this.onMatchViewportResolutionCallback = () => { console.log("Resolution changed and match viewport resolution is turned on, did you forget to override this function?"); }
+
+        // set play for video
+        this.videoElement.onclick = () => {
+            if (this.videoElement.paused) {
+                this.videoElement.play();
+            }
+        }
```
hmuurine: Is autoplay without clicking supported? Needs `this.videoElement.muted = true` to work in Chrome and mobile browsers, but with that setting should work without a click. Some use cases might require automatic muted start without needing to click the video.

Frontend/library/src/Config/Config.ts
```diff
@@ -0,0 +1,479 @@
...
+		const psSettingsHeader = document.createElement("div");
+		psSettingsHeader.classList.add("settingsHeader");
+		psSettingsHeader.classList.add("settings-text");
+		psSettingsHeader.textContent = sectionHeading;
+
+		// add section and heading to parent settings element
+		sectionElem.appendChild(psSettingsHeader);
+		settingsElem.appendChild(sectionElem);
+		return sectionElem;
+	}
```
hmuurine: Would it make sense to decouple the code that stores and modifies config values from the user interface code that is used to modify the values? Now these are coupled, but maybe some developers would like to implement their own user interface e.g. using React, Vue, Svelte etc. In that case it would make sense to store only the config values somewhere and have a separate user interface mechanism that just uses the config class methods to get values and update them.

Frontend/library/src/Config/SettingFlag.ts
```diff
@@ -0,0 +1,148 @@
+// Copyright Epic Games, Inc. All Rights Reserved.
+
+import { SettingBase } from "./SettingBase";
+
+export class SettingFlag extends SettingBase {
+
+    /* We toggle this checkbox to reflect the value of our setting's boolean flag. */
+    _checkbox : HTMLInputElement; // input type="checkbox"
```
hmuurine: Also same suggestion here and in other config data types: Could maybe decouple the flag boolean value from the user interface checkbox. So in config data structure there would be only booleans, and a separate UI layer could then use the config in whatever way they wish to render checkboxes, toggle buttons etc.

hmuurine (commented):
Great to see the JS monolith split into more manageable parts! Left a couple of suggestions as an initial review just based on reading through the code. Also looking forward to taking this for a spin and testing what the developer experience feels like.

Frontend/package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Frontend",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {}
+}
```
sheepsteak: I think this file can be deleted. There is no `package.json` in this folder.

lukehb: Good catch!

Frontend/library/src/AFK/AFKOverlay.ts
lukehb: Good call, I'll use some auto formatter I think. Too many devs touching this code, I think it is time to enforce some styling.

Frontend/library/src/Inputs/KeyboardController.ts
lukehb: Should we suspend sending all datachannel events to the UE application you think? Or just intercept the input events like keyboard, mouse, gamepad, hmd etc?

Frontend/library/src/VideoPlayer/VideoPlayer.ts
lukehb: Yeah we do support autoplaying without a click, so you are quite right we will need to start muted (though in many cases we actually send the audio track seperately and play it through an audio element instead of through the video, which achieves better latency in cases where they don't have to be perfectly synced).

Frontend/library/src/Config/Config.ts
lukehb: Yes, it would make sense! In general, I would like to move all the UI into its own library/plugin that can optionally used it the user wants to have the reference UI or otherwise provide their own.

Frontend/library/src/Config/SettingFlag.ts
lukehb: Agreed, as above, the whole lib needs a pass to decouple UI and library functionality so UI can be byo.

Frontend/implementations/EpicGames/tsconfig.json
lukehb: Fixed in: https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/a67b483a85b04ba02623513b67422e6a01af6c67

Frontend/library/src/VideoPlayer/VideoPlayer.ts
lukehb: On consideration we do make the user always do an initial "click to start" in our implementation, so we can also get away with just calling play without muting. If we did remove that though, we would need this.

Frontend/package-lock.json
lukehb: Fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/9dfc8e7140d5e9710892d5070eb4176da3233c5c

Frontend/library/src/AFK/AFKController.ts
lukehb: FIxed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/b600b7920e31324bfc0c2182b13ee91ea2097b06

Frontend/library/src/AFK/AFKController.ts
lukehb: Fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/b600b7920e31324bfc0c2182b13ee91ea2097b06

Frontend/library/src/AFK/AFKController.ts
lukehb: Fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/b600b7920e31324bfc0c2182b13ee91ea2097b06

Frontend/library/src/AFK/AFKOverlay.ts
lukehb: Fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/63/commits/b600b7920e31324bfc0c2182b13ee91ea2097b06

Frontend/library/src/Inputs/KeyboardController.ts
hmuurine: That's a good question. I would imaging that maybe input events are the more likely candidate. E.g. there could be a need to send some custom payload over the data channel based on what the user clicks in the HTML modal, so probably full data channel blocking might not be ideal.

Frontend/library/src/VideoPlayer/VideoPlayer.ts
hmuurine: If a developer uses this as a library, they might need the config option to enable `muted = true` for the video element already on init. I could imagine use cases where an application should start on page load without user interaction.

Frontend/library/package.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "devDependencies": {
+        "@typescript-eslint/eslint-plugin": "^5.16.0",
+        "@typescript-eslint/parser": "^5.16.0",
+        "cspell": "^4.1.0",
+        "eslint": "^8.11.0",
+        "prettier": "2.8.3",
+        "ts-loader": "^9.4.2",
+        "typedoc": "^0.23.24",
+        "webpack-cli": "^5.0.1"
```
hmuurine, line 24: Needs to add `webpack` and `typescript` as a devDependency. Tested to build this, and got an error about missing `webpack` and `typescript` if you don't have them installed globally.

Webpack-cli doesn't pull webpack as a transitive dependency since it's marked as a peerDependency, that needs to be provided by the project: https://github.com/webpack/webpack-cli/blob/master/package.json#L53

Tested and verified that the build passes with these installed as devDependencies.

```suggestion
        "typescript": "^4.9.4",
        "webpack": "^5.75.0",
        "webpack-cli": "^5.0.1"
```

Frontend/implementations/EpicGames/package.json
```diff
@@ -0,0 +1,24 @@
...
+  },
+  "devDependencies": {
+    "@epicgames/pixelstreamingfrontend": "^0.0.1",
+    "webpack-cli": "^5.0.1",
+    "webpack-dev-server": "^4.11.1",
+    "css-loader": "^6.7.3",
+    "html-loader": "^4.2.0",
+    "html-webpack-plugin": "^5.5.0",
+    "path": "^0.12.7",
+    "ts-loader": "^9.4.2",
```
hmuurine, line 19:
```suggestion
    "ts-loader": "^9.4.2",
    "typescript": "^4.9.4",
```
Needs typescript as a devDependency. The build fails unless you have typescript installed globally.

Frontend/library/src/pixelstreamingfrontend.ts
```diff
@@ -0,0 +1,57 @@
...
+export {
+    InboundAudioStats,
+    InboundVideoStats
+} from './PeerConnectionController/InboundRTPStats';
+export { OutBoundVideoStats } from './PeerConnectionController/OutBoundRTPStats';
+export { PixelStreamingApplicationStyle } from './Application/PixelStreamingApplicationStyles';
+
+import { PixelStreamingApplicationStyle } from './Application/PixelStreamingApplicationStyles';
+export const PixelStreamingApplicationStyles =
+    new PixelStreamingApplicationStyle();
```
hmuurine, line 57: The constructor of `PixelStreamingApplicationStyle` auto-applies the CSS styles, so this `new PixelStreamingApplicationStyle();` call auto-applies the styles whenever importing this file for the first time. If the users need custom styles, might want to not auto-apply the styles, but instead apply them explicitly when the application loads. Or maybe provide the style provider as part of the config, so could be customized?

lukehb: Agreed, that is fair, we should make the style be configurable or not auto apply.

hmuurine (approved):
Decoupling the default UI from the business logic and providing the functionality as a library might not be in the scope of this PR, so maybe those should be left for the future. The main thing here was Typescript conversion, which looks good to me! 👍 

(Maybe could reword the title of this PR before merge?)

Frontend/library/src/Config/SettingFlag.ts
lukehb: We have decided to move the decouple to separate task.

Frontend/library/src/Config/Config.ts
lukehb: We will do this in a separate task where we tackle decoupling UI from business logic.

Frontend/library/src/pixelstreamingfrontend.ts
lukehb: How we handle styling and its application to the reference frontend will be a done in separate task.

SignallingWebServer/platform_scripts/bash/setup.sh
```diff
@@ -91,18 +91,20 @@ function check_and_install() { #dep_name #get_version_string #version_min #insta
 function setup_frontend() {
 	# navigate to root
 	pushd ${BASH_LOCATION}/../../.. > /dev/null
-
+	export PATH="../../SignallingWebServer/platform_scripts/bash/node/bin:$PATH"
```
lukehb, line 94: @Belchy06 Will this keep prepending this on subsequent runs?

Belchy06: According to this [article](https://phoenixnap.com/kb/linux-add-to-path), adding a directory to the PATH variable only affects the current terminal session so should be good for multiple re-runs!

lukehb:
As @hmuurine mentioned, there is additional work of decoupling the UI from the core library. This is a separate issue - as tracked here: https://github.com/EpicGames/PixelStreamingInfrastructure/issues/66

Users should expect that decoupling work to land later (some time before (or a bit after) UE 5.2).

PR #67, opened 2023-02-07
Multi Streamer Support

mcottontensor:
Adding the ability to stream multiple streamers into the one signalling server.


Frontend/library/src/Config/SettingSelect.ts
```diff
@@ -0,0 +1,139 @@
+// Copyright Epic Games, Inc. All Rights Reserved.
```
Belchy06: Do we still need this file if we've merged to using the SettingOption?

SignallingWebServer/cirrus.js
```diff
@@ -270,387 +270,501 @@ if (config.UseHTTPS) {
...
 	}
-	const playerName = playerId == SFUPlayerId ? "SFU" : `player ${playerId}`;
-	const rawMsg = JSON.stringify(msg);
-	logOutgoing(playerName, msg.type, rawMsg);
-	player.ws.send(rawMsg);
+
+	if (!streamers.has(streamer.id)) {
+		console.error(`Disconnecting streamer ${streamer.id} does not exist.`);
+	} else {
+		sendStreamerDisconnectedToMatchmaker();
```
lukehb, line 391: Have these changes been tested against the matchmaker?

Frontend/library/src/Config/SettingSelect.ts
mcottontensor: No

SignallingWebServer/cirrus.js
mcottontensor: They have not. I wouldn't even know how to test the MM at this point

lukehb: We will focus of automated testing after this release so we don't have to manually do this stuff. But testing MM atm is like:

- Run the .bat in the MatchMaker folder
- Run 1 Cirrus with UseMatchMaker turned on
- Run another Cirrus with UseMatchMaker turned on and different http, ws, and sfu ports, e.g. 81, 7788, 8899
- Run a UE streamer and connect to Cirrus 1, e.g. 8888
- Run another UE streamer and connect to Cirrus 2, e.g. 7788
- Go to browser localhost:80 and localhost:81

Belchy06: Just amend last step, you'd wanna go to port 90 as that's the default matchmaker port.

Matchmaker should redirect you to one of the two free sessions.

PR #68, opened 2023-02-07
Handle DefaultToHover as a config option in InitialSettings

StomyPX:
UE_5-2 will pass DefaultToHover using a new ConfigOptions field in InitialSettings


lukehb (approved):
lgtm

PR #74, opened 2023-02-07
Multi streamers

mcottontensor:
Adding in support for multiple streamers.
- Streamers now have IDs to identify them.
- When connecting streamers will be requested to identify using the existing 'identify' message.
- When players connect to a signalling server with only one streamer, they will be automatically subscribed to that streamer to mimic old behaviour.
- When players connect to a signalling server with multiple streamers they will be required to select a streamer to subscribe to by using the dropdown in the player configuration menu.
- Players can specify a streamer to subscribe to using the URL parameter 'StreamerId' ie. ?StreamerId=MainStreamer which will auto subscribe to MainStreamer on connect.
- Players cannot yet switch streamers mid stream and will be required to reload the browser.

### DIFF
--- a/SignallingWebServer/cirrus.js
+++ b/SignallingWebServer/cirrus.js
@@ -790,7 +790,7 @@ if (config.UseMatchmaker) {
 			type: 'connect',
 			address: typeof serverPublicIp === 'undefined' ? '127.0.0.1' : serverPublicIp,
 			port: httpPort,
-			ready: streamer && streamer.readyState === 1,
+			ready: streamers.size > 0,
 			playerConnected: playerConnected
 		};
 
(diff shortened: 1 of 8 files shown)


SignallingWebServer/cirrus.js
```diff
@@ -270,387 +270,501 @@ if (config.UseHTTPS) {
...
+		} else {
+			streamer.ws.send(msgString);
+		}
+	}
+
+	sendTo(message) {
+		const msgString = JSON.stringify(message);
+		this.ws.send(msgString);
+	}
+};
```
lukehb, line 324: I know this class is using the streamers global, but wdyt about moving this class and the PlayerType object into its own file and passing streamers as param to constructor?

mcottontensor: Honestly there's a lot of things I'd like to do differently but it mostly falls into the realm of the rewrite/refactor we're hoping to do soon. So I'd be happy to do it but feel like it'd just be replaced when the inevitable refactor happens.

lukehb: Fair

PR #75, opened 2023-02-08
Indicate to the signalling server when the browser intends to send the offer

Belchy06:
This is useful when using UE5 as by default if the application receives a connection it will send an offer. In the case where the browser intends to send the offer we will connect but tell UE not to send the offer and instead wait to receive it.


lukehb (approved):
lgtm

PR #77, opened 2023-02-09
Fix: delayed mic input 

Belchy06:
Caused due to the frontend signalling stereo mic when mic infact mono


lukehb (approved):
lgtm - I think the issue was how this was handled on the UE side too, but this fixes the delay issue, so landing!

PR #83, opened 2023-02-09
Fix: Preferred codec selector causes no stream if application launched without `-PixelStreamingNegotiateCodecs`

Belchy06:
This was due to the codec preference array only containing a single (the preferred) codec. 

Now, we set the preferred codec to be the first in the array with the remaining browser supported codecs coming after.


lukehb (approved):
Perfect

PR #87, opened 2023-02-12
Multi streamers

mcottontensor:
This allows the front end to switch between streams at will using the drop down menu in settings.
I also added a small hint to the front end when you connect to a signalling server with multiple streamers so it doesn't feel like an error.


lukehb:
With https://github.com/EpicGames/PixelStreamingInfrastructure/commit/4fc43bcf2936af9c7e5bbff16f35cb466d4aec82 landed should this be closed now?

PR #88, opened 2023-02-13
Multi-streamer QOL improvements

Belchy06:
This PR adds a few QOL improvements to the library, specifically for the case where the are multiple streamers.

- The streamer selector is now updated on stream start to show you which streamer has been selected
- The PreferSFU toggle is now functional again
- Changing the streamer selector during a stream destroys the current peer connection and connects to the newly selected streamer (UE doesn't support re-negotiation so this is currently one of the only options)
-  Auto subscription on only one streamer has moved from cirrus to the frontend (tested and maintains backwards compatability).

The frontend now has an order of operations when checking which streamer to subscribe to. It is as follows:
- If there's a single streamer, subscribe to it regardless of what's in the URL
- If there's an SFU **AND** the PreferSFU toggle is on, subscribe to it regardless of what's in the URL
- If there's a streamer ID in the URL **AND** this streamer is connected to the signalling server, subscribe to it
- If none of the above conditions are met, no stream is subscribed to


Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1219,10 +1212,19 @@ export class WebRtcPlayerController {
...
         );
         messageStreamerList.ids.unshift(''); // add an empty option at the top
-        this.config.setOptionSettingOptions(
-            OptionParameters.StreamerId,
-            messageStreamerList.ids
-        );
+        this.config.setOptionSettingOptions(OptionParameters.StreamerId, messageStreamerList.ids);
+
+        const urlParams = new URLSearchParams(window.location.search);
+        if(messageStreamerList.ids.length == 2) {
```
lukehb: single streamer case is when length == 2, is that right?

Belchy06: Heh, yes because we add a blank option a few lines above. So in this case the two ids are ['', 'DefaultStreamer']

SignallingWebServer/Public/stresstest.html
```diff
@@ -1,42 +0,0 @@
-<!-- Copyright Epic Games, Inc. All Rights Reserved. -->
```
lukehb, line 1: Do we need a .gitignore for these?

Belchy06: SignallingWebServer/Public is already in the gitignore, but I have a feeling that these files were already tracked but the gitignore was updated without removing these files from being tracked.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: Might be worth making this clearer. ie. checking before we add the blank option. I can see this confusing someone later and changing it to break.

SignallingWebServer/cirrus.js
```diff
@@ -678,6 +679,7 @@ playerServer.on('connection', function (ws, req) {
 	var url = require('url');
 	const parsedUrl = url.parse(req.url);
 	const urlParams = new URLSearchParams(parsedUrl.search);
+	const browserSendOffer = urlParams.has('OfferToReceive') && urlParams.get('OfferToReceive') !== 'false';
```
lukehb, line 682: Did we always send this info to Cirrus or is this new on this commit?

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
lukehb: ^ Agreed

SignallingWebServer/cirrus.js
Belchy06: This info was added [here](https://github.com/EpicGames/PixelStreamingInfrastructure/commit/8be3b840b73df340ff30ef10aa343bc802d4970c) but squashed in the following PR. Just adding it back

lukehb: Yep. Ty.

SignallingWebServer/Public/stresstest.html
lukehb: Ah okay.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
Belchy06: Fair point. Fixed

lukehb (approved):
All commented have be addressed. Landing.

PR #91, opened 2023-02-14
Removing the player id from forwarded messages when they're sent to t…

mcottontensor:
Removing the player id field off forwarded messages to players.


lukehb (approved):
lgtm

PR #92, opened 2023-02-15
Bug fixes

Belchy06:
- Fix: Library not building on Linux due to incorrect casing
- Fix: Browsers that don't contain the xr object on the navigator wouldn't load the webpage
- QOL: Display a message when multiple streamers are detected, notifying users to select the stream they want


Belchy06:
Closing this as the PR was instead split into multiple others.

PR #94, opened 2023-02-15
QOL: Display a message when multiple streamers are detected

Belchy06:
notifying users to select the stream they want


lukehb (approved):
Approved

PR #97, opened 2023-02-15
Fix SFU data channels not being created

Belchy06:


lukehb (approved):
My bad, should've seen this error in the first one.

PR #101, opened 2023-02-17
Gracefully handle unsupported message types

Belchy06:
Fixed by not force closing the websocket


lukehb (approved):
All good. @Belchy06 can you ensure this fix is backported to the other branches too. (And forward ported to master/5.2 if needed)

PR #106, opened 2023-02-20
Neaten up install scripts

Belchy06:
Removed the custom install scripts from `Frontend/implementations/EpicGames/package.json` as this would override the default installer which led to no packages being installed.

Updated platform scripts to properly reflect this change.


lukehb (approved):
Let's land this

PR #117, opened 2023-02-22
Smaller bundle size

hmuurine:
# Summary

This PR makes the build bundle sizes smaller. There were a couple of issues that increased the size of both the final implementation `player.js` build size and the size of the library JS packages. The main issue was that the `library` code was included twice in the JS packages.

## Bundle size of `library` and `ui-library`

The library builds bundled all their external and internal dependencies within the `dist/*.js` package.
- Could be excluded by adding `externals` section in library Webpack builds
- The only one who needs to import and bundle these external dependencies is the final implementation application
- `dist/*.js` size reductions:
   - `library` 128kB -> 118kB
   - `ui-library` 217kB -> 67kB (previously bundled also `library` within the JS in addition to the external dependencies)

## Bundle size of `player.js`

The build pipeline bundled two versions of the library code into the final package. This could be seen by opening and investigating the final `player.js`, which had two versions of the library code.
- Both the local build and the NPM version of `library` were added to the bundle
- Local build was added because the `implementation/EpicGames` linked locally to it via `npm link ../library`
- NPM version was added because `ui-library` ended up using the NPM version when the final build was made. Reasons:
   - Running `npm ci` after `npm link` overwrites the link
   - Also running first `npm link ../library` in `ui-library`, but then running `npm link ../../ui-library` in `EpicGames` resets the linking in `ui-library`. The links need to be set up in opposite order if both need to be preserved: first in `EpicGames`, then in `ui-library`.
- Fixing this has the greatest effect: player.js 347k -> 220kB

## Exporting ESM modules

The library modules were exported as `umd` modules, which isn't tree-shakeable.
- Added `esm` module build in addition to `umd`
- `esm` module is provided in package.json `module` and `umd` in package.json `main`. Dependency users are automatically able to choose which ones to use
- The final build was able to shake something out of the code, but not much: player.js 220kB -> 218kB
   - Maybe expected since the sample application uses pretty much all the functionality that is available through the library

### DIFF
--- a/Frontend/library/package.json
+++ b/Frontend/library/package.json
@@ -3,7 +3,9 @@
     "version": "0.1.0",
     "description": "Frontend library for Pixel Streaming",
     "main": "dist/lib-pixelstreamingfrontend.js",
+    "module": "dist/lib-pixelstreamingfrontend.esm.js",
     "types": "types/pixelstreamingfrontend.d.ts",
+    "sideEffects": false,
     "scripts": {
         "compile": "tsc --build --clean && tsc",
         "build": "npx webpack --config webpack.prod.js",
--- a/Frontend/ui-library/package.json
+++ b/Frontend/ui-library/package.json
@@ -14,6 +14,7 @@
         "spellcheck": "cspell \"{README.md,.github/*.md,src/**/*.ts}\""
     },
     "devDependencies": {
+        "@epicgames-ps/lib-pixelstreamingfrontend-ue5.2": "0.1.0",
         "@typescript-eslint/eslint-plugin": "^5.16.0",
         "@typescript-eslint/parser": "^5.16.0",
         "cspell": "^4.1.0",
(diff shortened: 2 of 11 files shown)


Frontend/library/package.json
```diff
@@ -3,7 +3,9 @@
     "version": "0.1.0",
     "description": "Frontend library for Pixel Streaming",
     "main": "dist/lib-pixelstreamingfrontend.js",
+    "module": "dist/lib-pixelstreamingfrontend.esm.js",
     "types": "types/pixelstreamingfrontend.d.ts",
+    "sideEffects": false,
```
hmuurine, line 8: This just promises that our library does not do any side effects when the library is imported. All the effects are done through direct calls, not just by importing. This allows the tree shaking mechanisms to drop any library code that isn't referenced.

Frontend/ui-library/package.json
```diff
@@ -26,11 +29,13 @@
         "webpack-cli": "^5.0.1"
     },
     "dependencies": {
-        "@epicgames-ps/lib-pixelstreamingfrontend-ue5.2": "0.1.0",
         "jss": "^10.9.2",
         "jss-plugin-camel-case": "^10.9.2",
         "jss-plugin-global": "^10.9.2"
     },
+    "peerDependencies": {
+        "@epicgames-ps/lib-pixelstreamingfrontend-ue5.2": "0.1.0"
```
hmuurine, line 37: Require `library` as a peer dependency provided through the implementation application dependencies, not as a transitive dependency through `ui-library`. This might prevent us from having two distinct versions as dependencies in the final build, one from direct dependency, one through ui-lib.

lukehb (approved):
Great work! Landing.

PR #123, opened 2023-02-24
Fix cirrus Dockerfile

Belchy06:
Thanks to `Darb Dude#1324` from the Unreal Slackers community for reporting this one.

We should really get some tests in.


lukehb (approved):
Ah, we forgot to call the new setup scripts... doh! Landing.

PR #129, opened 2023-02-28
Use mediasoup's internal stream ids for SCTP to avoid clashes

### LINKED ISSUE #62: Game crashes with more than 2 connections when using PixelStreaming with SFU
With reference implementations of the Signalling server and the SFU, when trying to connect from the 3rd tab, the game crashes. This issue is not reproducible if ?preferSFU=false.

```
Signal 11 caught.
Malloc Size=262146 LargeMemoryPoolOffset=262162
CommonUnixCrashHandler: Signal=11
[2023.01.16-12.13.11:591][244]LogCore: === Critical error: ===
Unhandled Exception: SIGSEGV: invalid attempt to write memory at address 0x0000000000000003

[2023.01.16-12.13.11:591][244]LogCore: Assertion failed: RecvResult.ok() [File:.\../Plugins/Media/PixelStreaming/Source/PixelStreaming/Private/PixelStreamingDataChannel.cpp] [Line: 52]
Failed to create Data Channel. Msg=Data channel creation failed
```

StomyPX:
This should allow more than 2 peers to connect to the SFU without having Unreal Engine crash from stream channels being double-assigned.

Fixes #62

### DIFF
--- a/SFU/sfu_server.js
+++ b/SFU/sfu_server.js
@@ -44,30 +44,7 @@ async function onStreamerOffer(sdp) {
 }
 
 function getNextStreamerSCTPId() {
-  if(!streamer){
-    throw new TypeError('Cannot generate an SCTP stream id - streamer was null.');
-  }
-  if (!streamer.transport || !streamer.transport.sctpParameters || typeof streamer.transport.sctpParameters.MIS !== 'number') {
-    throw new TypeError('Streamer was not setup with the following require properties: streamer.transport.sctpParameters.MIS');
-  }
-  const numStreams = streamer.transport.sctpParameters.MIS;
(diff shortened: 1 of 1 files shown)


lukehb (commented):
Moving to using the internal allocation of stream ids that mediasoup is doing under the hood. Slightly risky using internal API, but I think we are okay with this for now if it fixes the issue.

@mcottontensor for your review also.

mcottontensor:
Looks alright to me.

lukehb (approved):
Landing!

lukehb:
This addresses: https://github.com/EpicGames/PixelStreamingInfrastructure/issues/62

PR #132, opened 2023-03-01
Api support for UE communication

hmuurine:
# Summary

Adds support for browser <-> UE communication through the API:
- `emitUIInteraction(data: object | string)`:
    - sends arbitrary data as an `UIInteraction` message
- `emitCommand(descriptor: object)`:
    - sends a Command message 
    - if the descriptor contains a `ConsoleCommand`, it is permitted only if the UE app has been started with`-AllowPixelStreamingCommands` command line argument
- `emitConsoleCommand(command: string)`:
    - sends a ConsoleCommand message 
    - permitted only if the UE app has been started with `-AllowPixelStreamingCommands` command line argument
- `addResponseEventListener(name, funct)` / `removeResponseEventListener(name)`:
    - register/remove a listener that is triggered when receiving UE -> browser messages

# Test plan

## Browser <-> UE
- Modified UE Pixel Streaming application to do the following:
   - Monitors for any incoming browser -> UE messages that contain the string `EchoBack`
   - If one received, send back a response with content `EchoResponse: $originalMessage`
blueprint
- Modified `uiless.ts` to do the following:
   - Sends a `UIInteraction` message starting with `EchoBack` 10 seconds after page has been opened
   - Logs any responses received from UE
Screenshot 2023-03-01 at 13 55 30
- Started uiless.html and waited for 10 seconds
- Verified that the browser -> UE and UE -> browser messages were both logged, ensuring that communication works both ways
Screenshot 2023-03-01 at 13 42 55

## Custom commands
- Started a Pixel Streaming UE app with command line parameter `-AllowPixelStreamingCommands`
- Added some code to `uiless.ts` to toggle FPS counter and degrade video quality 10 seconds after initialization:
```
setTimeout(() => {
	pixelStreaming.emitCommand({
		ConsoleCommand: "stat fps",
	});
	pixelStreaming.emitCommand({
		"Encoder.MinQP": 51,
	});
}, 10000);
```
- Opened `uiless.html`, waited for 10 seconds, and verified that FPS counter was enabled and video quality was degraded
Screenshot 2023-02-28 at 9 28 26
Screenshot 2023-03-02 at 9 05 52

- Repeated the test with `-AllowPixelStreamingCommands` command line arg disabled
- Verified that FPS counter was not enabled after 10 seconds, since the console command was rejected. Video was still degraded as expected

### DIFF
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.ts
@@ -566,7 +566,7 @@ export class PixelStreaming {
 
     /**
      * Send data to UE application. The data will be run through JSON.stringify() so e.g. strings
-     * and any serializable objects with no recurrence can be sent.
+     * and any serializable plain JSON objects with no recurrence can be sent.
      * @returns
      */
     public sendUIInteraction(descriptor: object) {
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.ts
@@ -579,15 +579,18 @@ export class PixelStreaming {
     }
 
     /**
-     * Send a console command to UE application. Only allowed if UE has signaled that it allows
-     * console commands.
+     * Send a command to UE application. Blocks ConsoleCommand descriptors unless UE
+     * has signaled that it allows console commands.
      * @returns true if succeeded, false if rejected
      */
-    public emitConsoleCommand(command: string) {
-        if (!this.allowConsoleCommands || !this._webRtcController.videoPlayer.isVideoReady()) {
+    public emitCommand(descriptor: object) {
+        if (!this._webRtcController.videoPlayer.isVideoReady()) {
             return false;
         }
-        this._webRtcController.emitConsoleCommand(command);
+        if (!this.allowConsoleCommands && 'ConsoleCommand' in descriptor) {
+            return false;
+        }
+        this._webRtcController.emitCommand(descriptor);
         return true;
     }
 
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.ts
@@ -53,6 +53,7 @@ export class PixelStreaming {
     private _videoElementParent: HTMLElement;
 
     _showActionOrErrorOnDisconnect = true;
+    private allowConsoleCommands = false;
 
     private onScreenKeyboardHelper: OnScreenKeyboard;
 
@@ -462,9 +463,9 @@ export class PixelStreaming {
             new InitialSettingsEvent({ settings })
         );
         if (settings.PixelStreamingSettings) {
-            const allowConsoleCommands =
-                settings.PixelStreamingSettings.AllowPixelStreamingCommands;
-            if (allowConsoleCommands === false) {
+            this.allowConsoleCommands =
+                settings.PixelStreamingSettings.AllowPixelStreamingCommands ?? false;
+            if (this.allowConsoleCommands === false) {
                 Logger.Info(
                     Logger.GetStackTrace(),
                     '-AllowPixelStreamingCommands=false, sending arbitrary console commands from browser to UE is disabled.'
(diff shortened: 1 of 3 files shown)


Frontend/library/src/UeInstanceMessage/SendDescriptorController.ts
```diff
@@ -45,7 +45,7 @@ export class SendDescriptorController {
      * @param messageType - UE Message Type
      * @param descriptor - Descriptor Message as JSON
      */
-    sendDescriptor(messageType: string, descriptor: object) {
+    sendDescriptor(messageType: string, descriptor: object | string) {
```
hmuurine, line 48: Supporting `object | string` to match what is promised in https://docs.unrealengine.com/5.1/en-US/customizing-the-player-web-page-in-unreal-engine/ 
> When you call the emitUIInteraction function, you can pass it a single string or JavaScript object.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1671,6 +1671,32 @@ export class WebRtcPlayerController {
...
+        Logger.Log(
+            Logger.GetStackTrace(),
+            '----   Sending custom Command:ConsoleCommand message   ----',
+            6
+        );
+        this.sendDescriptorController.emitCommand({
+            ConsoleCommand: command,
+        });
+    }
+
```
lukehb, line 1697: There is also another option we may as well expose here, which is `emitCommand(descriptor)` that calls 

```ts
this.sendDescriptorController.emitCommand(descriptor);
```

This is used for inbuilt commands we always accept (like changing EncoderMinQP for example) and newly in 5.2 we also expose on the C++ side a way for users to bind custom command messages so we should also expose a way for users to send these using our new library.

lukehb: For the test plan, on the C++ side you should be able to something like this to set a custom handler:

```C++
IPixelStreamingModule& PSModule = IPixelStreamingModule::Get();
PsModule.GetStreamer(PSModule.GetDefaultStreamerID())->GetInputHandler()->SetCommandHandler(YourHandler)`
```

Or you can test using one of the existing in-built commands like `Encoder.MinQP`

lukehb (changes requested):

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
hmuurine: Sounds good, if it's ok to expose this more capable interface. I'll make the changes 👍

hmuurine: Removed `emitConsoleCommand` and added `emitCommand` in commit https://github.com/EpicGames/PixelStreamingInfrastructure/pull/132/commits/70c14a5fc6b78f6fc24ca534dd4c08a0125c189d

Also the test description updated accordingly.

lukehb (approved):

Frontend/library/src/PixelStreaming/PixelStreaming.ts
```diff
@@ -564,6 +565,55 @@ export class PixelStreaming {
...
+        if (!this._webRtcController.videoPlayer.isVideoReady()) {
+            return false;
+        }
+        if (!this.allowConsoleCommands && 'ConsoleCommand' in descriptor) {
+            return false;
+        }
+        this._webRtcController.emitCommand(descriptor);
+        return true;
+    }
+
```
lukehb, line 596: @hmuurine Sorry to be a bother on this one. But on further consideration, we should have both `emitCommand` and `emitConsoleCommand`. Mostly keeply the latter to match existing 5.1 documentation.

lukehb (changes requested):

Frontend/library/src/PixelStreaming/PixelStreaming.ts
hmuurine: No worries, restored the previous implementation and now both are there. 👍

lukehb (approved):
lgtm!

PR #133, opened 2023-03-01
Customize styles through UI API

hmuurine:
# Summary

Added code that allows the users to customize styles via the UI API.

-   No longer applying styles automatically in constructor. Need to call `applyStyles()`
-    Developers can give custom overriding CSS styles as optional parameters in `new PixelStreamingApplicationStyle()` constructor
-    Moved light/dark mode palette applying code from `Application` to `PixelStreamingApplicationStyle` to keep all the style related code in a single place and separate from `Application`
- `Application` now exposes an optional `onUpdateColors` callback that is called when light/dark mode is toggled. The style class hooks into this to apply styles on mode change
-    Developers can give light/dark mode palette as an optional parameter in `new PixelStreamingApplicationStyle()` constructor to override the default colors

# Test plan

-    Ran the frontend without any style overrides, and verified that the default styles were applied
Screenshot 2023-03-01 at 17 12 42
-    Toggled between light/dark mode and verified that the styles were applied
Screenshot 2023-03-01 at 17 12 51
-    Modified player.ts to apply custom styles and verified that the styles were applied
Screenshot 2023-03-01 at 17 17 02
Screenshot 2023-03-01 at 17 17 32
-    Modified player.ts to set a custom light/dark mode palette and verified that the styles were applied on toggling color mode
Screenshot 2023-03-01 at 17 20 57
Screenshot 2023-03-01 at 17 20 19
Screenshot 2023-03-01 at 17 20 28

### DIFF
--- a/Frontend/ui-library/src/Application/Application.ts
+++ b/Frontend/ui-library/src/Application/Application.ts
@@ -27,7 +27,7 @@ import { ConfigUI, LightMode } from '../Config/ConfigUI';
 
 export interface UIOptions {
     stream: PixelStreaming;
-    onUpdateColors?: (isLightMode: boolean) => void;
+    onColorModeChanged?: (isLightMode: boolean) => void;
 }
 
 /**
@@ -56,14 +56,14 @@ export class Application {
 
     configUI: ConfigUI;
 
-    onUpdateColors: UIOptions["onUpdateColors"];
+    onColorModeChanged: UIOptions["onColorModeChanged"];
 
     /**
      * @param options - Initialization options
      */
     constructor(options: UIOptions) {
         this.stream = options.stream;
-        this.onUpdateColors = options.onUpdateColors;
+        this.onColorModeChanged = options.onColorModeChanged;
         this.configUI = new ConfigUI(this.stream.config);
 
         this.createOverlays();
@@ -88,7 +88,7 @@ export class Application {
 
         this.showConnectOrAutoConnectOverlays();
 
-        this.updateColors(this.configUI.isCustomFlagEnabled(LightMode));
+        this.setColorMode(this.configUI.isCustomFlagEnabled(LightMode));
     }
 
     public createOverlays(): void {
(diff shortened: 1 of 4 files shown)


Frontend/ui-library/src/Styles/PixelStreamingApplicationStyles.ts
```diff
@@ -1,493 +1,534 @@
...
+            width: '100%'
+        },
+        '#qualityStatus': {
+            fontSize: '37px',
+            paddingRight: '4px'
+        },
+        '.svgIcon': {
+            fill: 'var(--color2)'
         }
     };
```
hmuurine, line 519: The style diff looks a bit strange but there are zero changes to the styles. Just removed the `global` wrapper which is applied below when combining the default styles with the overrides. The changes above are caused by decreased indentation & prettier auto-formatting.

lukehb:
CC @Belchy06 for vis

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -27,6 +27,7 @@ import { ConfigUI, LightMode } from '../Config/ConfigUI';
 
 export interface UIOptions {
     stream: PixelStreaming;
+    onUpdateColors?: (isLightMode: boolean) => void;
```
lukehb: I know this PR is just following the existing naming convention of `updateColors`; however, given it is now exposed, I propose we rename that method and all the related methods to `setColorMode(isLightMode: boolean)` so we end up with `onColorModeChanged` or similar. Wdyt @hmuurine @Belchy06 ?

More generally though, I am happy with the changes.

lukehb (changes requested):

Frontend/ui-library/src/Application/Application.ts
hmuurine: Good idea! The new name describes better what it does. I updated the names in commit https://github.com/EpicGames/PixelStreamingInfrastructure/pull/133/commits/5c98f21073b0d54668ee5add95302d8acead04cd

lukehb (approved):
lgtm!

PR #136, opened 2023-03-02
Enable/disable user input devices

hmuurine:
# Summary

Added configuration for enabling & disabling the following user input devices:
- Keyboard
- Mouse
- Touch
- Game pad
- XR controller

The config can be given either as initial configuration when creating the `Config` object, or can be changed on the fly. The changed config is applied immediately by removing/reapplying the event handlers related to user input.

Added the new configuration items also to the sample UI so that they can be toggled on demand.

The event unregistering is done via a new utility class `EventListenerTracker` that keeps track of the registered event handlers and can execute unregister commands for them when needed. This same mechanism can be used later if implementing teardown functionality that unregisters everything.

# Test plan

## Configuration via API
- Started a new Pixel Streaming session and verified that mouse/keyboard/touch/game pad input were all enabled by default and worked as before
- Modified `player.ts` to disable all user input on init
Screenshot 2023-03-02 at 16 20 34
- Verified that now the UI did not respond to any mouse/keyboard/touch/game pad input when connected

## Dynamic run-time config change
- Started a new session with all user input enabled
- Enabled & disabled mouse/keyboard/touch/game pad input one by one and verified that each configuration change enabled/disabled the corresponding input device as expected

https://user-images.githubusercontent.com/8232602/222460844-40a694cd-84e5-49a9-bc9a-74ba50cb5f34.mp4


Frontend/library/src/Inputs/FakeTouchController.ts
```diff
@@ -140,7 +163,7 @@ export class FakeTouchController implements ITouchController {
                         x - this.fakeTouchFinger.x,
                         y - this.fakeTouchFinger.y
                     );
-                toStreamerHandlers.get('MoveMouse')([
+                toStreamerHandlers.get('MouseMove')([
```
hmuurine, line 166: This was a bug in master: There is no handler named `MoveMouse`, so it failed if hitting this row.

Frontend/library/src/Inputs/GamepadController.ts
```diff
@@ -185,7 +211,9 @@ export class GamePadController {
             }
             this.controllers[controllerIndex].prevState = currentState;
         }
-        this.requestAnimationFrame(() => this.updateStatus());
```
hmuurine, line 188: This failed in master version because `this` should point to `window` when calling `requestAnimationFrame`. Fixed by calling `bind(window)` at row 29.

Frontend/library/src/Inputs/InputClassesFactory.ts
```diff
@@ -61,110 +58,29 @@ export class InputClassesFactory {
...
-        videoElementParent.onwheel = (wheelEvent: WheelEvent) =>
-            hoveringMouseEvents.handleMouseWheel(wheelEvent);
-        videoElementParent.ondblclick = (mouseEvent: MouseEvent) =>
-            hoveringMouseEvents.handleMouseDouble(mouseEvent);
-        videoElementParent.pressMouseButtons = (mouseEvent: MouseEvent) =>
-            hoveringMouseEvents.handlePressMouseButtons(mouseEvent);
-        videoElementParent.releaseMouseButtons = (mouseEvent: MouseEvent) =>
-            hoveringMouseEvents.handleReleaseMouseButtons(mouseEvent);
-    }
-
```
hmuurine, line 167: Moved these two functions to `MouseController` to keep all mouse handling code in a single place.

Frontend/library/src/VideoPlayer/VideoPlayer.ts
```diff
@@ -8,8 +8,6 @@ import { Logger } from '../Logger/Logger';
  */
 declare global {
     interface HTMLElement {
-        pressMouseButtons?(mouseEvent: MouseEvent): void;
-        releaseMouseButtons?(mouseEvent: MouseEvent): void;
```
hmuurine, line 12: Removed these and the corresponding function handlers that were appended to videoParent since it turned out they were never used. Maybe some legacy code?

hmuurine:
Notice if someone wants to test Gamepad functionality, at least the demo UE app sent an invalid `Protocol` message that had the wrong `messageLength` for `GamepadAnalog` input (had length 3, but should be 10). To test game pad support properly, I needed to overwrite `GamepadAnalog` length by value 10 after receiving the wrong value from UE. Did not submit that into this PR since it's an UE side issue.

Might be just an issue in my test application though, which is just something grabbed from UE samples. Maybe some newer test programs send a proper value and don't need the hack.

lukehb:
@Belchy06 To review this one.

Also Will can you look into the gamepad issue reported here.

lukehb (changes requested):
Should we also do the same for XR inputs?

Frontend/library/src/Config/Config.ts
```diff
@@ -398,6 +402,50 @@ export class Config {
...
+            Flags.EnableGamePadInput,
+            new SettingFlag(
+                Flags.EnableGamePadInput,
+                'Enable game pad input',
+                'If enabled, send game pad events to streamer',
+                true,
+                useUrlParams
+            )
+        );
+
```
lukehb, line 460: This page will also need updating with these additions: https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/Frontend/Docs/Settings%20Panel.md

Frontend/library/src/Inputs/FakeTouchController.ts
lukehb: @Belchy06 for vis

hmuurine:
> Should we also do the same for XR inputs?

I did look at this yesterday, but the XR input seems to work in a slightly different way compared to the rest of the input methods. The code doesn't register any event listeners. It's instead requesting a XR session, and when the session is active, the code has access to XR controller state. So I'm not 100% sure it fits into this same pattern.

Maybe we could add the XR flag in Config, and if that's disabled, block sending the controller update messages to UE. Wouldn't work the same way under the hood, but looking from the dev point of view could be nice to have that flag available.

hmuurine:
XR flag implemented in commit https://github.com/EpicGames/PixelStreamingInfrastructure/pull/136/commits/71daadfd34529efdab89e72dfece9b87bee2a207

Frontend/library/src/Config/Config.ts
hmuurine: Docs have been updated 👍

Frontend/library/src/Inputs/GamepadController.ts
Belchy06: Can confirm I got this issue with the old front end and it is now fixed. Thank you!

Belchy06:
Looks good. I just made one change to simplifying the flag names from EnableDeviceInput to DeviceInput as user's should be able to get the idea of enabling/disabling from the toggle itself.

I was unable to repro the issue you brought up with the analogue gamepad input, with mine 'just working'. Which samples were you using and I'll look to seeing if it need to be fixed?

hmuurine:
> I was unable to repro the issue you brought up with the analogue gamepad input, with mine 'just working'. Which samples were you using and I'll look to seeing if it need to be fixed?

I was using a sample I found at the Samples tab in Epic Games Store.
egs-ps-demo

Just loaded that sample into UE, built a Windows executable and started it with Pixel Streaming command line arguments. Didn't make any code changes to the sample application. Also used UE version 5.1 if that makes a difference.

Belchy06:
Actually an engine level issue https://github.com/EpicGames/UnrealEngine/blob/5.1/Engine/Plugins/Media/PixelStreaming/Source/PixelStreaming/Private/PixelStreamingModule.cpp#L526 😬

lukehb (approved):

lukehb:
re: gamepad issue - @Belchy06 I'll leave it up to you to open a ticket UE side if needed.

PR #137, opened 2023-03-02
Fix datachannels when using an SFU

mcottontensor:
This fixes the issue where only the last joined peer could interact with the given streamer so now all peers should be able to interact by default.
The problem was that the new signalling server was making some assumptions about messages being sent from players and would unconditionally add the playerId field to the message with the sending player. Unfortunately the streamerDataChannels message used that field to specify the id of the player we were creating the data channel for. Without this fix, the data channel would always get the player id of the SFU and not the requesting player.

### DIFF
--- a/SignallingWebServer/cirrus.js
+++ b/SignallingWebServer/cirrus.js
@@ -307,6 +307,8 @@ class Player {
 			return;
 		}
 
+		// normally we want to indicate what player this message came from
+		// but in some instances we might already have set this (streamerDataChannels) due to poor choices
 		if (!message.playerId) {
 			message.playerId = this.id;
 		}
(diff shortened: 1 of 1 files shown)


SignallingWebServer/cirrus.js
```diff
@@ -307,7 +307,9 @@ class Player {
 			return;
 		}
 
-		message.playerId = this.id;
+		if (!message.playerId) {
```
lukehb, line 312: Behaviour is quite nuanced, I feel a comment is necessary here saying something about player id might be already set in some cases (such as sfu datachannels) so we don't want to blanket overwrite it

lukehb (changes requested):

lukehb (approved):

PR #141, opened 2023-03-07
Clean up on Pixel Streaming session disconnect

hmuurine:
# Summary

The Pixel Streaming library should clean up after it has finished a session. This helps integrating the lib within a single-page-app like a React app, where you might have the user navigate from a Pixel Streaming view to other views without a page refresh.

The current code leaves e.g. dangling event handlers connected to `videoParent` and `document` element, which it should clean up when not required any more. One of the symptoms of these dangling event handlers is a lot of console logging after disconnected whenever mouse or keyboard events are emitted:
Screenshot 2023-03-07 at 12 26 49

This PR performs the following clean up tasks when the library has closed the session:
- Unregister all user input event handlers when disconnecting
   - These are registered again always on reconnecting, so it makes it symmetric if unregistering on disconnecting
- Use `clearTimeout`, not `clearInterval` to clear AFK timeout since it's not an interval
   - (actually in some browsers `clearInterval` accidentally clears also timeouts if called with a timeout id because of underlying implementation details, but this is not something guaranteed for all browsers)
- Added `exitPointerLock()` call to pointer lock mouse handler teardown
   - Exits pointer lock when disconnecting, but also if the lock happens to be active when some external API call toggles mouse input off or switches to hovering mouse event handler
- Also made `pixelStreaming.disconnect()` more robust by adding `undefined` checks
   - It previously crashed if `disconnect()` was called before connection was established. Got hit by this when implementing a test React app if navigating quickly out of the Pixel Streaming view before the connection was fully up

# Test Plan

## AFK timetout
- Opened a Pixel Streaming session with AFK timeout enabled
- Disconnected the UE app before the AFK timeout was triggered
- Waited and verified that the timeout was not triggered after disconnected

## Unregister event handlers on disconnecting
- Opened a Pixel Streaming session in `uiless.html`
- Listed the event handlers connected to `videoParentElement` and `document`, and saw a bunch of event handlers
Screenshot 2023-03-07 at 12 50 07
- Used the API to auto-disconnect after 10 seconds
Screenshot 2023-03-07 at 12 51 26
- After disconnected listed the event handlers again and verified that all had been cleared
Screenshot 2023-03-07 at 12 50 35
- Moved mouse around and pressed keyboard keys. Verified that there were no more console messages printed by the event handlers

- Retested the above by disconnecting from UE side, not through API. Verified that the event handlers were cleared also on UE side disconnect

- Used `player.html` to verify that disconnecting & reconnecting multiple times re-established the event handlers every time the connection was established. Verified that user input worked as expected after multiple reconnections

## Pointer lock
- Opened a Pixel Streaming session with pointer lock mouse handler
- Clicked the video to lock mouse cursor
- Disconnected the UE app to trigger disconnect actions
- Verified that the pointer lock was released when disconnecting

### DIFF
--- a/Frontend/library/src/AFK/AFKController.ts
+++ b/Frontend/library/src/AFK/AFKController.ts
@@ -76,7 +76,7 @@ export class AFKController {
     stopAfkWarningTimer() {
         this.active = false;
         this.countdownActive = false;
-        clearInterval(this.warnTimer);
+        clearTimeout(this.warnTimer);
         clearInterval(this.countDownTimer);
     }
 
--- a/Frontend/library/src/Inputs/MouseController.ts
+++ b/Frontend/library/src/Inputs/MouseController.ts
@@ -134,6 +134,15 @@ export class MouseController {
         this.mouseEventListenerTracker.addUnregisterCallback(
             () => lockedMouseEvents.unregisterMouseEvents()
         );
+        this.mouseEventListenerTracker.addUnregisterCallback(() => {
+            if (
+                document.exitPointerLock &&
+                (document.pointerLockElement === videoElementParent ||
+                    document.mozPointerLockElement === videoElementParent)
+            ) {
+                document.exitPointerLock();
+            }
+        });
     }
 
     /**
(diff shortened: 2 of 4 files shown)


lukehb (approved):
lgtm!

PR #149, opened 2023-03-13
Pass HTTPS certificate locations via Cirrus configuration

marcinbiegun:
This change adds two new fields to cirrus configuration file: `HTTPSCertFile` and `HTTPSKeyFile`. Those are relative file locations of certificate files used to establish secure HTTPS connection by the Cirrus server.

Thanks to this change I'm able to keep certificate files outside of `PixelStreamingInfrastructure` project location.

The default values for the new fields point to previous default certificate files locations, so if the new parameters are not provided, the app haves the same way as before.


lukehb (approved):
Love it. Merged!

PR #151, opened 2023-03-14
Enable SFU layer preference messages from UE

Belchy06:
Currently, the SFU will auto switch the layer based on the prevailing network conditions. This is great for production use, but visually inspecting what each layer looks like is tedious and in its current state requires the use of a third party tool such as NetLimiter to restrict the browsers bitrate and trigger a layer switch.

This PR adds a `layerPreference` message type use to trigger a layer switch manually.

If Cirrus receives a message of `layerPreference` type, it pass it straight to the SFU. The SFU then extracts the `playerId` from the message and sets the target spatial and temporal layers for this player.


lukehb:
This PR idea was abandoned?

PR #161, opened 2023-03-17
Added ability to optionally disable certain frontend elements

kass-kass:
# Problem statement:

The frontend classes (mainly `Application` and `Controls`) are set to automatically create elements, such as:
 - Full screen button
 - Settings button + panel
 - Stats button + panel
 - Video QP indicator
 - XR button

However, for many applications, it might be beneficial to allow disabling certain functionalities. For example, an end-user-facing web page with a stream might be better off allowing for much settings changes, as well as simplifying the experience for the user by not displaying technologically advanced information.

Also, the styling and location of the buttons is pre-defined, and it's not easily possible to use other HTML elements from a custom implementation of the frontend. 

# Solutions

The changes in this pull request address these limitations by expanding the `UIOptions` interface used to initialize an `Application` by including optional parameters to configure:
 - Which panels (settings, stats) should (not) be created
 - Which buttons should be created, and if they're not created, if they're totally disabled, or instead use an externally provided `HTMLButtonElement` to trigger the same functionality
 - Whether or not the video QP indicator should be created internally, or an external element should be used. The connection to the external element is a callback function which simply provides the qp value, which can be handled in any way necessary.

# Compatibility

The changes are made in a way that is fully compatible with the previous API: all the new configuration parameters are optional, and are checked whenever they might be used. If no changes are made to the code interacting with UI library, no visible or functional changes to the behaviour will be present. By default, if no custom config is specified, all elements are created in the same way as before.

# Details: 
The `Application` configuration object is updated, and the class also creates a newly-added configuration object for the `Controls` class, which now checks which buttons it should create.

The `FullScreenIcon` class is now split into a `FullScreenIconBase` base class with all fullscreen toggling functionality, which can be triggered by any provided button; a `FullScreenIconExternal` class, which initializes any given button element to have fullscreen toggle functionality; and `FullScreenIcon`, which retains the previous functionality it had - creating a button element with svg-based icons and changing its state internally.


Frontend/ui-library/src/UI/FullscreenIcon.ts
```diff
@@ -66,19 +67,91 @@ export class FullScreenIcon {
...
+    constructor(externalButton : HTMLButtonElement) {
+        super();
+        this.rootElement = externalButton;
+    }
+
+}
+
+/**
+ * An implementation of FullScreenIconBase that a button and associated visuals
+ * (svgs for each state) internally.
```
lukehb: > An implementation of FullScreenIconBase that a button and associated visuals (svgs for each state) internally.

to

> The default fullscreen icon that contains a button and svgs for each state.

Frontend/ui-library/src/UI/FullscreenIcon.ts
```diff
@@ -66,19 +67,91 @@ export class FullScreenIcon {
     }
 
     /**
-     * Get the the button containing minimize and maximize.
+     * Makes the document fullscreen
```
lukehb: > Makes the document fullscreen

to

> Makes the document or fullscreenElement fullscreen.

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -550,7 +629,21 @@ export class Application {
      * @param QP - the quality number of the stream
      */
     onVideoEncoderAvgQP(QP: number) {
-        this.videoQpIndicator.updateQpTooltip(QP);
+        // If no custom config was provided, we have an internal indicator
+        if (this._options.videoQpUiConfig == undefined) this.videoQpIndicator.updateQpTooltip(QP);
```
lukehb: The rest of the codebase uses braces for `if/else` can you modify this to match.

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -550,7 +629,21 @@ export class Application {
...
+
+        else switch(this._options.videoQpUiConfig.indicatorType) {
+            case (UIElementType.CreateElement) : {
+                this.videoQpIndicator.updateQpTooltip(QP);
+                break;
+            }
+            case (UIElementType.UseExternalElement) : {
+                let callback = this._options.videoQpUiConfig.externalIndicatorUpdateCallback;
+                if (!!callback) callback(QP);
```
lukehb: Missing braces on `if`

Frontend/ui-library/src/UI/Controls.ts
```diff
@@ -5,6 +5,21 @@ import { SettingsIcon } from './SettingsIcon';
...
+    statsButtonType? : UIElementType,
+    fullscreenButtonType? : UIElementType,
+    settingsButtonType? : UIElementType,
+    xrIconType? : UIElementType
+}
+
+function shouldCreateButton(type : UIElementType | undefined) : boolean {
+    if (type == undefined) return true;
+    else return type === UIElementType.CreateElement;
+}
```
lukehb, line 26: Missing braces on `if/else`

Frontend/ui-library/src/UI/Controls.ts
```diff
@@ -20,11 +35,15 @@ export class Controls {
...
+    constructor(config? : ControlsUIConfiguration) {
+        if (!config || shouldCreateButton(config.statsButtonType))
+            this.statsIcon = new StatsIcon();
+        if (!config || shouldCreateButton(config.fullscreenButtonType))
+            this.settingsIcon = new SettingsIcon();
+        if (!config || shouldCreateButton(config.settingsButtonType))
+            this.fullscreenIcon = new FullScreenIcon();
+        if (!config || shouldCreateButton(config.xrIconType))
+            this.xrIcon = new XRIcon();
     }
```
lukehb, line 55: Missing braces on `if`

Frontend/ui-library/src/UI/Controls.ts
```diff
@@ -34,10 +53,10 @@ export class Controls {
...
             this._rootElement.id = 'controls';
-            this._rootElement.appendChild(this.fullscreenIcon.rootElement);
-            this._rootElement.appendChild(this.settingsIcon.rootElement);
-            this._rootElement.appendChild(this.statsIcon.rootElement);
-            WebXRController.isSessionSupported('immersive-vr').then(
+            if (!!this.fullscreenIcon) this._rootElement.appendChild(this.fullscreenIcon.rootElement);
+            if (!!this.settingsIcon) this._rootElement.appendChild(this.settingsIcon.rootElement);
+            if (!!this.statsIcon) this._rootElement.appendChild(this.statsIcon.rootElement);
+            if (!!this.xrIcon) WebXRController.isSessionSupported('immersive-vr').then(
                 (supported: boolean) => {
```
lukehb, line 75: Missing braces on `if`

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
+/** Whether a stream UI element is internally made, externally provided, or disabled. */
+export enum UIElementType {
+    CreateElement,
```
lukehb: I would prefer `CreateElement` be called `CreateDefaultElement`

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
...
+    UseExternalElement,
```
lukehb: I would prefer `UseExternalElement` be called `CreateCustomElement`

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
+/** Whether a stream UI element is internally made, externally provided, or disabled. */
+export enum UIElementType {
```
lukehb: I would prefer this enum be called `UIElementCreationMode`

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
+/** Whether a stream UI element is internally made, externally provided, or disabled. */
```
lukehb, line 1: Enum description will need to updated slightly if the other changes are made.

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
...
+export enum UIElementType {
+    CreateElement,
+    UseExternalElement,
+    Disable
+}
+
+/**
+ * This is used to define settings for a stream-related UI panel:
+ * is it created, and if it is, what kind of button is used to show/hide it.
+ */
```
lukehb, line 23: >  * This is used to define settings for a stream-related UI panel: is it created, and if it is, what kind of button is used to show/hide it.

to

> Configures how a general stream-related UI panel. For example: is it created, and if it is, what kind of button is used to show/hide it. This configuration is used for the settings panel and stats panel by default.

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
```diff
@@ -0,0 +1,23 @@
...
+export type PanelConfiguration = {
+    // If panel is enabled, HTML elements for it will be created, and funtionality bound
+    isEnabled : boolean,
+    // (Only relevant if isEnabled) The type of the button to show/hide this panel
+    visbilityToggleButtonType? : UIElementType
+    // Only relevant if visbilityToggleButtonType is UseExternalElement
+    externalButton? : HTMLButtonElement | undefined
```
lukehb: I would like propose another class be created

`UIElementConfig`

that contains:


- `UIElementCreationMode`
- `SetCustomElement()` or something to that effect so user can pass whatever custom HTMLElement they so desire

Then in the `PanelConfiguration` there is a `visibilityBtnConfig: UIElementConfig`

Frontend/ui-library/src/UI/Controls.ts
```diff
@@ -5,6 +5,21 @@ import { SettingsIcon } from './SettingsIcon';
...
 import { WebXRController } from '@epicgames-ps/lib-pixelstreamingfrontend-ue5.2';
+import { UIElementType } from '../UI/UIConfigurationTypes'
+
+
+export type ControlsUIConfiguration = {
+    //[Property in keyof Controls as `${Property}Type`]? : UIElementType;
+    statsButtonType? : UIElementType,
+    fullscreenButtonType? : UIElementType,
+    settingsButtonType? : UIElementType,
+    xrIconType? : UIElementType
```
lukehb: Similar to below I would like these to be

```ts
statsButtonConfig? : UIElementConfig,
fullscreenButtonConfig? : UIElementConfig,
settingsButtonConfig? : UIElementConfig,
xrIconConfig? : UIElementConfig
```

See below comment for explanation of `UIElementConfig`

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -550,7 +629,21 @@ export class Application {
...
+        // If no custom config was provided, we have an internal indicator
+        if (this._options.videoQpUiConfig == undefined) this.videoQpIndicator.updateQpTooltip(QP);
+
+        else switch(this._options.videoQpUiConfig.indicatorType) {
+            case (UIElementType.CreateElement) : {
+                this.videoQpIndicator.updateQpTooltip(QP);
+                break;
+            }
+            case (UIElementType.UseExternalElement) : {
+                let callback = this._options.videoQpUiConfig.externalIndicatorUpdateCallback;
```
lukehb: I would prefer this the ui library expose an event emitter for the videoQP update that can be subscribed to externally.

Similar to how we do in the core library with: https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/Frontend/library/src/Util/EventEmitter.ts

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -18,20 +18,56 @@ import { PlayOverlay } from '../Overlay/PlayOverlay';
...
+      * If needed, this behaviour can be configured. */
+    settingsPanelConfig?: PanelConfiguration;
+    /** By default, a stats panel and associate visibility toggle button will be made.
+      * If needed, this behaviour can be configured. */
+    statsPanelConfig?: PanelConfiguration;
+    /** By default, a video QP indicator element will be made.
+      * If needed, this behaviour can be configured. */
+    videoQpUiConfig? : VideoQpIndicatorConfiguration,
+    /** If needed, the full screen button can be external or disabled. */
+    fullScreenButtonType? : UIElementType,
```
lukehb: I would prefer fullscreen and xr fields be removed here in favour of some sort of ControlsConfiguration (see my below suggestions for use of `UIElementConfig` types).

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -18,20 +18,56 @@ import { PlayOverlay } from '../Overlay/PlayOverlay';
...
 import { SettingsPanel } from '../UI/SettingsPanel';
 import { StatsPanel } from '../UI/StatsPanel';
 import { VideoQpIndicator } from '../UI/VideoQpIndicator';
 import { ConfigUI, LightMode } from '../Config/ConfigUI';
+import { UIElementType, PanelConfiguration, isPanelEnabled } from '../UI/UIConfigurationTypes'
+import { FullScreenIconBase, FullScreenIconExternal } from '../UI/FullscreenIcon';
+
+type VideoQpIndicatorConfiguration = {
+    // Is the indicator created internally, or should an external element be used?
+    indicatorType : UIElementType
```
lukehb: Move to `UIElementConfig` type as per my below suggestions

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -18,20 +18,56 @@ import { PlayOverlay } from '../Overlay/PlayOverlay';
...
+    /** If the indicatorType is UseExternalElement, this function will be called with the
+      * QP (quantization parameter) value. The lower the QP, the better the video quality.
+      * The range of the QP value varies per codec: 1 to 127 for VP8, 1 to 51 for H.264. */
+    externalIndicatorUpdateCallback? : (qp : number) => void
```
lukehb: Move to an event emitter as per my below suggestions.

lukehb (changes requested):
Very nice PR. I have left my comments about what I would like to changed before I can approve. In general, my changes are to bring it in line with coding standards in the rest of the codebase and to make the feature proposed here even more flexible (I hope).

lukehb:
@hmuurine To review also if you have any free cycles.

kass-kass:
@lukehb Thank you for the detailed feedback, will make these changes in the coming few days.

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
kass-kass: I don't understand why it should be called "create" custom element? I assume the element is already created elsewhere, and the PS classes will simply use the provided element. Therefore `UseCustomElement` makes more sense to me, but maybe I misunderstand the intention for this?

In my new commits I renamed `UseExternalElement` to `UseCustomElement`.

Frontend/ui-library/src/Application/Application.ts
kass-kass: In cases where we don't want to create an internal video QP indicator, I guess it might be easier to just listen directly to the stream's event with `stream.addEventListener('videoEncoderAvgQP', ...)`. Therefore I changed the video QP indiactor configuration to a simple boolean create/don't create, and added as a comment to that boolean this event's description for convenience.

Frontend/ui-library/src/Application/Application.ts
kass-kass: I understand that including a `ControlsUIConfiguration` directly in the `UIOptions` of the `Application` is concise, but it would duplicate the stats/settings button configs, which are currently stored in the panel configurations. For logical consistency, we need to keep the only one config per button.

It's possible to remove the button configs from the panel configurations (that would mean we can create a panel only if the button is enabled), but that isn't a great solution for two reasons: 
- It results in an overall less flexible approach, as it limits creating a panel to cases where a button is made/used. It might make sense in some situations to create the panels but not buttons, for example if someone wants to display stats/settings permanently docked to the side.
- It makes things more confusing. It's good to have the `PanelConfiguration` contain both `isEnabled` and the associated `visibilityButtonConfig`, as they are logically interlinked. Otherwise it's not obvious where an associated control element for a panel is defined. 

Therefore for now I kept the two separate panel configs, and additional `UIElementConfigs` for buttons that have no panels. The `UIOptions` interface did become more concise in any case, since the XR and fullscreen configurations each went from two variables to one.

Let me know if this is an issue.

kass-kass:
Hi @lukehb, added the changes that you suggested and resolved most comments. I think I updated everything you mentioend overall, some things I did slightly differently but not in any significant way. See the replies to the comments that I left.

https://github.com/EpicGames/PixelStreamingInfrastructure/pull/161/commits/fd46393529291dd3128b07aeddcb1a70d680d2bb

Frontend/ui-library/src/UI/UIConfigurationTypes.ts
lukehb: Fair comment. Happy with how you renamed it.

Frontend/ui-library/src/Application/Application.ts
lukehb: I'm happy with how your latest commit ended up doing it.

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -49,19 +46,14 @@ export interface UIOptions {
...
-    xrExternalButton? : HTMLButtonElement
-}
-
-function shouldCreateVideoQpIndicator(config : VideoQpIndicatorConfiguration | undefined) {
-    return !config || (!!config && config.indicatorType === UIElementType.CreateElement);
+    xrControlsConfig? : UIElementConfig,
+    /** By default, a video QP indicator element will be made. If needed, this can be disabled.
+     *  For custom UI elements to react to the QP being changed, use a PixelStreaming object
+     *  and addEventListener('videoEncoderAvgQP', ...) or removeEventListener(...). */
+    disableInternalVideoQpIndicator? : boolean,
```
lukehb: I liked this being a VideoQPIndicatorConfig object as opposed to a boolean - for both consistency and extensibility reasons. For example, consider that, we may, at some later point, add more config options to the qp indicator. Whereas now, if we wanted to do, we would break public API if we introduced this boolean then removed it in favour of a config type later. I propose you reintroduce the config type you had before except called `VideoQPIndicatorConfig`. I know it only has one boolean atm, but I think you understand my reasoning as explained above.

lukehb (changes requested):
@kass-kass Left one minor request for further changes, everything is looking good otherwise.

Frontend/ui-library/src/Application/Application.ts
kass-kass: Yes, sounds good! Made the change :)

kass-kass:
@lukehb Should all be there, thanks for the explanation.

lukehb (changes requested):
Code changes look great. Ready to merge on that front... but...

Sorry to be that guy, but we just recently made a push to add more documentation customizing the frontend, so given this change I was wondering whether your PR could also include a new section on [this page](https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/Frontend/Docs/Customizing%20the%20Player%20Webpage.md) explaining the new functionality you have added here.

Specifically if you could document:

- How to turn off the various UI elements this PR enables
- How to provide custom elements

Very much appreciated if you could do this! After that, let's land this... for real this time :)

kass-kass:
@lukehb Totally reasonable, it's good to have things well described. Will update the documentation.

kass-kass:
@lukehb Added explanations for these features and code snippets.

Additionally made some small changes to structure of the customization doc, and also made some links in the main readme file to make separate projects directly reachable with a link.

lukehb (approved):
Amazing work. Merged!

PR #163, opened 2023-03-19
Fixing security warnings.

mcottontensor:
Addressing:
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/1
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/2
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/3
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/4
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/5
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/6
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/7
https://github.com/EpicGames/PixelStreamingInfrastructure/security/code-scanning/8

In most cases the recommended action was taken which resulted in a few extra required packages.


SignallingWebServer/cirrus.js
```diff
@@ -78,8 +79,10 @@ if (config.UseFrontend) {
 	var httpPort = 3000;
 	var httpsPort = 8000;
 
-	//Required for self signed certs otherwise just get an error back when sending request to frontend see https://stackoverflow.com/a/35633993
-	process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
+	if (config.DisableSSLCert) {
+		//Required for self signed certs otherwise just get an error back when sending request to frontend see https://stackoverflow.com/a/35633993
+		process.env.NODE_TLS_REJECT_UNAUTHORIZED = "0"
+	}
```
lukehb, line 86: @MWillWallT @Belchy06 When testing VR Pixel Streaming with local SSL cert this will mean this flag/config will have to be set. Meaning there is a small update required to the docs.

SignallingWebServer/cirrus.js
```diff
@@ -189,6 +192,16 @@ if (config.UseHTTPS) {
 
 sendGameSessionData();
 
+// set up rate limiter: maximum of five requests per minute
+var RateLimit = require('express-rate-limit');
+var limiter = RateLimit({
+  windowMs: 1*60*1000, // 1 minute
+  max: 5
```
lukehb: Let's bump this to like 60 requests per minute, 5 might actually have trouble serving some of our content

SignallingWebServer/cirrus.js
Belchy06: Could we somehow get this flag combined with the `UseHTTPS` one? eg if using HTTPS then allow self signed certs by default?

lukehb (changes requested):
lgtm, made minor comment about bumping rate limit

SignallingWebServer/cirrus.js
lukehb: @mcottontensor Just had a chat with @Belchy06 about this one. In the above `if` statement can we please add a big warning log when you start Cirrus and that codepath gets hit, something like:

> WARNING: config.DisableSSLCert is true. Unauthorized SSL certificates will be allowed! This is convenient for local testing but please DO NOT SHIP THIS IN PRODUCTION. To remove this warning please set DisableSSLCert to false in your config.json.

And then additionally, @Belchy06 would like the `config.json` file to have `DisableSSLCert` set to `true` by default so it is convenient for local testing in the VR usecase.

lukehb (approved):

PR #165, opened 2023-03-20
Fix: controller indices from multiple peers would clash

### LINKED ISSUE #164: Gamepad indices clash with multiple peers
## Problem
When multiple peers are connected, each peer's controllers will increment from 0.

This means that if player 0 has 2 gamepads, they'll be assigned index 0 and index 1 and when player 1 connects with two gamepads of its own, they'll also be assigned index 0 and index 1 as illustrated below:

Blank diagram (1)

The clashing of these gamepads prevents use cases such as local multiplayer (co-op) experiences.

Belchy06:
## Problem
When multiple peers are connected, each peer's controllers will increment from 0.

This means that if player 0 has 2 gamepads, they'll be assigned index 0 and index 1 and when player 1 connects with two gamepads of its own, they'll also be assigned index 0 and index 1 as illustrated below:

image

## Solution
The most suitable solution to this problem would be for the Pixel Streaming module to keep track of all of the in-use gamepad indices. When a new gamepad is detected by the peer, it sends a request to the application for a gamepad index that is free. The application responds with an index to use and the peer will then use this provided controller index for any subsequent calls to the application.

## Compatibility
Backwards compatibility is preserved in this change by using the local controller index (the one that increments from 0) if the application doesn't provide the peer with an index. This case would happen if using a UE version less than 5.2 but using the latest frontend.

### DIFF
--- a/Frontend/library/src/DataChannel/InitialSettings.ts
+++ b/Frontend/library/src/DataChannel/InitialSettings.ts
@@ -30,7 +30,6 @@ export class InitialSettings {
 export class PixelStreamingSettings {
     AllowPixelStreamingCommands?: boolean;
     DisableLatencyTest?: boolean;
-	PlayerId?: number;
 }
 
 /**
--- a/Frontend/library/src/DataChannel/InitialSettings.ts
+++ b/Frontend/library/src/DataChannel/InitialSettings.ts
@@ -30,6 +30,7 @@ export class InitialSettings {
 export class PixelStreamingSettings {
     AllowPixelStreamingCommands?: boolean;
     DisableLatencyTest?: boolean;
+	PlayerId?: number;
 }
 
 /**
--- a/Frontend/library/src/UeInstanceMessage/StreamMessageController.ts
+++ b/Frontend/library/src/UeInstanceMessage/StreamMessageController.ts
@@ -176,6 +176,11 @@ export class StreamMessageController {
             structure: ['uint8', 'uint16', 'uint16', 'uint8', 'uint8', 'uint8']
         });
         // Gamepad Input Messages. Range = 90..99
+        this.toStreamerMessages.add('GamepadConnected', {
+            id: 93,
+            byteLength: 0,
+            structure: []
+        });
         this.toStreamerMessages.add('GamepadButtonPressed', {
             id: 90,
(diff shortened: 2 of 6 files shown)


lukehb:
Update PR with a link to the UE side commit when it lands too - as the two are both required for the solutions to be complete.

Frontend/library/src/UeInstanceMessage/StreamMessageController.ts
```diff
@@ -176,6 +176,11 @@ export class StreamMessageController {
             structure: ['uint8', 'uint16', 'uint16', 'uint8', 'uint8', 'uint8']
         });
         // Gamepad Input Messages. Range = 90..99
+        this.toStreamerMessages.add('GamepadConnected', {
+            id: 93,
+            byteLength: 0,
+            structure: []
```
lukehb, line 182: No controller id passed on gamepad connected?

lukehb: This is because id is generated from UE side not browser side, right?

lukehb:
Have you tested this with multiple repeated controller plug/unplug attempts?

Frontend/library/src/UeInstanceMessage/StreamMessageController.ts
Belchy06: Correct

Frontend/library/src/DataChannel/InitialSettings.ts
```diff
@@ -30,6 +30,7 @@ export class InitialSettings {
 export class PixelStreamingSettings {
     AllowPixelStreamingCommands?: boolean;
     DisableLatencyTest?: boolean;
+	PlayerId?: number;
```
lukehb: With this `PlayerId` is this always sent with initial settings now from UE side?

lukehb: Can you point me at where this is used elsewhere in this commit?

Belchy06: Was left over from testing. Now removed

lukehb (approved):
All feedback addressed. Merging.

Issue #167, opened 2023-03-20
[BUG] - NPM emits warnings when running Jest + Bundled NodeJS used in platform_scripts


lukehb:
**UE Version: **
Any

**Frontend Version: **
Since f58b32cae6c92a3886cfc5f3a088cc21aa82ecea

**Problem component**
- [x] Signalling Server
- [x] Frontend library

**Description**
When running the unit tests in the frontend library using `npm run test` I am presented with the following warning log output from NPM/Jest:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@jest/console@29.5.0',
npm WARN EBADENGINE   required: { node: '^14.15.0 || ^16.10.0 || >=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.4.2', npm: '7.18.1' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@jest/core@29.5.0',
npm WARN EBADENGINE   required: { node: '^14.15.0 || ^16.10.0 || >=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.4.2', npm: '7.18.1' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@jest/environment@29.5.0',
npm WARN EBADENGINE   required: { node: '^14.15.0 || ^16.10.0 || >=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.4.2', npm: '7.18.1' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@jest/expect@29.5.0',
npm WARN EBADENGINE   required: { node: '^14.15.0 || ^16.10.0 || >=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.4.2', npm: '7.18.1' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@jest/expect-utils@29.5.0',
npm WARN EBADENGINE   required: { node: '^14.15.0 || ^16.10.0 || >=18.0.0' },
npm WARN EBADENGINE   current: { node: 'v16.4.2', npm: '7.18.1' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
[...]

PR #168, opened 2023-03-20
Downgrade ts-jest and jest to support older Node version

hmuurine:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:

The script in `platform_scripts/cmd/run_local.bat` installs NodeJS version 16.4.2, which is not compatible with `ts-jest` version 28.0.0 or newer. https://github.com/EpicGames/PixelStreamingInfrastructure/commit/f58b32cae6c92a3886cfc5f3a088cc21aa82ecea installed `ts-jest` 29.0.5, which resulted in `Unsupported engine` errors if using NodeJS 16.4.2. See issue: https://github.com/EpicGames/PixelStreamingInfrastructure/issues/167

The versions of all Jest packages need to be downgraded to 27.x.y series, which are compatible with NodeJS version 16.4.2.

## Solution

Downgraded `ts-jest` package to version 27.1.5, which is the latest version that supports Node 16.4.2. Also downgraded `jest` and `jest-environment-jsdom` to the latest 27 series version to fulfill `ts-jest` 27 requirements.

## Documentation

No changes in documentation.

Ran `npm install` and `npm run test` with NodeJS 16.4.2 and verified that there were no longer complaints about NodeJS version mismatch. Verified that the unit tests worked like before with the downgraded Jest version.


lukehb (approved):
Lgtm!

PR #172, opened 2023-03-22
Fix: sfu player would try subscribing when sfu disconnected

### LINKED ISSUE #171: [BUG] - onSFUDisconnected sfuPlayer.Subscribe()
**UE Version: **
UE 5.2

**Problem component**
SignallingWebServer/cirrus.js

**Description**
in the function onSFUDisconnected, sfuPlayer uses subscribe when it is ready to disconnect. It looks like a clerical error because it makes no sense to do so.

**Expected behavior**
I think it should be unsubscribe instead of subscribe.

Belchy06:
## Relevant components:
- [x] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
In Cirrus, when the SFU disconnects, the player representing the SFU would try and subscribe to the SFU. This is just a simple type, the player should instead unsubscribe.

## Solution
Fix the type. The player now unsubscribes on SFU disconnect.

Fixes https://github.com/EpicGames/PixelStreamingInfrastructure/issues/171

Tested streaming both with and without the SFU. Checked that on SFU disconnect, the player will unsubscribe instead of subscribing.

### DIFF
--- a/SignallingWebServer/cirrus.js
+++ b/SignallingWebServer/cirrus.js
@@ -561,7 +561,7 @@ function onSFUDisconnected() {
 	disconnectAllPlayers(SFUPlayerId);
 	const sfuPlayer = getSFU();
 	if (sfuPlayer) {
-		sfuPlayer.subscribe();
+		sfuPlayer.unsubscribe();
 		sfuPlayer.ws.close(4000, "SFU Disconnected");
 	}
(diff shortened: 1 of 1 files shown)


lukehb (approved):
lgtm

PR #179, opened 2023-03-28
Add support for programmatically changing peer layers when using the SFU

Belchy06:
## Relevant components:
- [X] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [X] SFU

## Problem statement:
Currently when utilising the SFU, players are switched to different resolution streams based on their prevailing network conditions.

It may be useful however, to force a specific player (or all players) to be on a specific layer. This could be useful during development to see how your stream will look at different resolutions without needing to use a third party tool such as NetLimiter to fake this poor network.

## Solution
This PR adds support for a new message from the application. This message is structured as follows:
```
{
    "type": "layerPreference",
    "playerId": "TargetPlayerId",
    "spatialLayer": N,
    "temporalLayer": M
}
```
The `playerId` value represents the id of the player whose layers you wish to change.

The `spatialLayer` value represents the resolution based layer, configured by the `SimulcastParameters` passed to the application at launch. If no value is passed, layer 0 represents full resolution and layer 1 represents half resolution.

The `temporalLayer` value represents the framerate based layer. Pixel Streaming currently doesn't support different temporal layers, but if/when support for it is added this value will become useful.

## Documentation
No SFU specific docs.

Tested using the latest SFU and Cirrus to ensure a stream can still operate as expected. 

I also setup a key bind in an Unreal project to switch between layers. In this example, I passed three layers in the application by launching with `-SimulcastParameters=1.0,5000000,20000000,2.0,1000000,5000000,4.0,50000,1000000`. These three layers represent Full Res, Half Res and Quarter Res. My bind would then cycle through each layer on key press and set every player to this new layer.

image

### DIFF
--- a/SFU/sfu_server.js
+++ b/SFU/sfu_server.js
@@ -44,7 +44,7 @@ async function onStreamerOffer(sdp) {
 }
 
 function getNextStreamerSCTPId() {
-	return streamer.transport._getNextSctpStreamId();
+    return streamer.transport._getNextSctpStreamId();
 }
 
 function onStreamerDisconnected() {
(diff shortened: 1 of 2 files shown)


lukehb (approved):
Approved

lukehb:
@mcottontensor This change should be merged back down in 5.2 branch and will mean a minor version bump to the RELEASE_VERSION file and we should also do a new release of the signalling server container too.

Issue #160, opened 2023-03-17
Connecting to Unreal 5.1 app with the 5.2 frontend crashes on connect


kass-kass:
Using the npm package `lib-pixelstreamingfrontend-ue5.2` with the standard signalling server and an Unreal app (both UE 5.1), and connecting to WebSocket connection on localhost, clicking the "play" button on the frontend, the streamer app crashes with
```
LogPixelStreaming: Error: Failed to set remote description: Failed to set remote answer sdp: Called in wrong state: stable
LogWindows: Error: appError called: Assertion failed: IsInGameThread() || IsInSlateThread() || IsInAsyncLoadingThread() [File:D:\build\++UE5\Sync\Engine\Source\Runtime\Slate\Public\Framework\Application\SlateApplication.h] [Line: 255] 
```

According to the debugger it comes from `FPixelStreamingInputHandler::HandleOnMouseLeave`, which triggers `FSlateApplication::Get()`. 

I understand that the readme of PixelStreamingInfrastructure states
>⚠️ There are breaking changes between UE versions - so make sure you get the right version. ⚠️

so this might be expected, since it's UE5.2 frontend + UE5.1 application. However, the latest stable engine version available is 5.1, and the only currently available frontend package is `-5.2`, so this is the only combination I could get.

Belchy06:
Hi @kass-kass,

Thanks for creating this issue!

I tried repro-ing today using UE5.1.1 and the library you mentioned above and was unfortunately unable to repro, using both browser send offer and application send offer.

Are you able to share with us (either through public repo or adding us as maintainers to a private repo) your implementation?

Cheers

kass-kass:
@Belchy06 Will try to isolate the issue today  in an empty project and send some more details on what exactly triggers the crash.

kass-kass:
@Belchy06, 

## Crash demonstration repository:
https://github.com/kass-kass/PixelStreamCrashDemo. The repo contains:
 - a bare-bones Unreal 5.1 project (PS plugin enabled, a single map created with a text moving up and down).
 - a bare-bones implementation of the PS infra (a ts/webpack project, with several lines of code in index.ts which create an `Application` object to display/manage it)

## Steps to reproduce:
1. Prepare webpage 
    1. Go to the TS project root folder and `npm install .`, then `npm run build`
    2. Open the generated `/dist/index.html` page
2. Prepare streamer
    1. Open project in Unreal Engine 5.1.1 editor
[...]

PR #180, opened 2023-03-28
Fix: Connecting to Unreal 5.1 app with the 5.2 frontend crashes on connect

Belchy06:
## Relevant components:
- [X] Signalling server
- [ ] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
This PR addresses https://github.com/EpicGames/PixelStreamingInfrastructure/issues/160. Using a UE version < 5.2 with latest infra would cause a crash.

## Solution
The root cause of the issue was that older versions of UE would always send offer on player connected, and player connected is caused by subscription which triggered by receiving a streamer list.

The solution in this case is to only request a streamer list if UE will be the one to send the offer. This does have the added side effect of the frontend not being able to work with multiple streamers when the browsers sends the offer.

This unfortunate drawback is required for cases such as Scalable Pixel Streaming where if we tried to request the streamer list, we would be met with no response and therefore never be able to subscribe.

Tested with UE5.1 and 5.2 and on both versions testing with BrowserSendOffer both on and off. I also tested multi-streamer still worked with UE sending offer on 5.2

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -200,7 +200,13 @@ export class WebRtcPlayerController {
             this.setVideoEncoderAvgQP(0);
         };
         this.webSocketController.onOpen.addEventListener('open', () => {
-            this.webSocketController.requestStreamerList();
+            const BrowserSendsOffer = this.config.isFlagEnabled(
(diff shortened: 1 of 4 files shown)


lukehb (approved):
lgtm - if we want multistreamer and browser send offer we can tackle this later - but I think this will require signalling clients who need that to support list streamer basically

PR #191, opened 2023-04-06
Security updates for default turn server configuration.

mcottontensor:
## Relevant components:
- [x] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
The default turn server configuration has a serious security vulnerability allowing internal network traversal from an external entity.

## Solution
This change introduces a set of configuration options in a config file that will close off traversal to certain subnets.

## Documentation
NA

Running the existing with-turn or turn scripts on windows/linux/docker should now mention that it loaded a config file and dump out a list of the blocked subnets.

### DIFF
--- a/SignallingWebServer/platform_scripts/bash/Dockerfile
+++ b/SignallingWebServer/platform_scripts/bash/Dockerfile
@@ -6,6 +6,8 @@ FROM node:latest
 # Copy the signalling server source code to the Docker build context
 COPY . /opt/SignallingWebServer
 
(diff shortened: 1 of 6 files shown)


lukehb:
Changes look okay - passing over to @MWillWallT to verify if this still works as expected on AWS Linux/Windows.

lukehb:
Fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/196

PR #192, opened 2023-04-07
Added ability to mute/unmute mic and enable it later than PixelStreaming creation time

kass-kass:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Currently, if a microphone track is enabled at `PixelStreaming` instance creation time, there is no easily accessible way to mute the microphone stream.

Additionally, if the `PixelStreaming` instance is created without a mic track, there is no easy way to add one later.

## Solution
This PR adds public funcitons to set the mic muted state.

The function to unmute the mic can be called with a `forceEnable` parameter, which allows to add a mic track at a later time (if one wasn't initially added), and trigger a full reconnect.

## Documentation
https://github.com/EpicGames/PixelStreamingInfrastructure/commit/252b958bb7507419b5fcb7371d46535a26304944

Doesn't modify existing APIs, just adds new functions.

Note: some intenral functionality that was very verbose and hard-to-read is now encapsulated in a new `RTCUtils` class.


Frontend/library/src/PeerConnectionController/PeerConnectionController.ts
```diff
@@ -426,18 +422,16 @@ export class PeerConnectionController {
             });
         } else {
             // set the audio options based on mic usage
-            const audioOptions = useMic
```
kass-kass, line 428: To explain this change: the `useMic ? ` check is extraneous, as this code is inside an `if (!useMic) ... else { }` scope.

Belchy06: Good catch!

Belchy06:
Looking good @kass-kass.

We're currently winding down active development of the master branch in preparation for the 5.2 launch, meaning we're only merging critical bug fixes.

We'll open back up to features once 5.2 has launched.

With that being said, I've got a small feature request to go with this PR. Could I get you to add a mic toggle in the top left corner (where fullscreen and panel buttons sit)? I think having it only be visible if the page is loaded with the `useMic` URLParam or if the `Use Mic` toggle is set, but once connected being able to click this button to trigger the `pixelStreaming` objects `muteMicrophone` and `unmuteMicrophone` functionality.

mcottontensor:
Is the extra functionality re. the microphone button coming? I have been looking at this PR and noticed the functionality was missing.

lukehb:
@mcottontensor  I think we can assume not. You can add it if you think the effort is low enough, otherwise I am happy to merge without the added UI button.

kass-kass:
@lukehb @mcottontensor Haven't been able to find the time to add this yet, due to other workloads. Will likely take a look beginning of October to finish this up, does that sound OK?

lukehb:
@mcottontensor I think we can merge without the mic button - the lack of that button is not blocking and tbh I prefer the feature without it.

mcottontensor (approved):

PR #193, opened 2023-04-07
Added exports for UI configuration types

kass-kass:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Some types which I added before weren't exported in the module.ts file, which was making it impossible to use them when the module is imported from npm.

## Solution
They are now exported.

## Documentation
None needed.

None needed.


lukehb (approved):
lgtm

PR #195, opened 2023-04-11
Alter reconnection flow to request streamer message list and fail out after N attempts.

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
When disconnected from a stream, the frontend would make no attempt to reconnect. This causes Virtual Cameras to never reconnect when changing target viewport.

To test this yourself (using 5.2):
1) Add 2 VCam Actors
2) Give both Streams their respective IDs, eg Stream 1 and Stream 2 (keep note of which is which)
3) Ensure both are set to Target Viewport 1 (as they would be by default)
4) Connect tab one to Stream 1
5) Connect a new tab Stream 2
6) Change Stream 2 to target Viewport 2 (do not disable VCamActor first)

You'll notice that Stream 2 will temporarily stop and then restart when it changes viewport.

## Solution
This issue has been addressed by altering web socket disconnects to enable a reconnection flow. When a web socket connection is closed by way of a streamer disconnecting, the frontend will reconnect to the signalling server and request the streamer list a maximum of N times to see if the streamer it was previously connected to has come back. In this case, N can be configured with `MaxReconnectAttempts` URL parameter or appropriate option in the settings panel.

If the streamer comes back within the N attempts, the frontend will automatically subscribe. Otherwise, if the streamer doesn't reconnect, the frontend will close the web socket connection and a user will be taken back to the original 'Click To Start' landing page.

## Documentation
Updated the Settings Panel document to reflect the new option.

Added a unit test to test the new reconnection flow.

### DIFF
--- a/Frontend/Docs/Settings Panel.md
+++ b/Frontend/Docs/Settings Panel.md
@@ -26,7 +26,7 @@ This page will be updated with new features and commands as they become availabl
 | **Suppress browser keys** | Suppress or allow certain keys we use in UE, for example F5 to show shader complexity instead of refreshing the page. |
 | **AFK if Idle** | Timeout the connection if no input is detected for a period of time. |
 | **AFK timeout** | Allows you to specify the AFK timeout period. |
-
+| **Max Reconnects** | The maximum number of reconnects the application will attempt when a streamer disconnects. |
 
 ### UI
 | **Setting** | **Description** |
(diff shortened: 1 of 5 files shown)


lukehb (approved):
Approved, I will leave this up to you if this change needs to make it into the 5.2 branch and a new release for 5.2 cut?

PR #215, opened 2023-05-03
Fix: Reconnects will be attempted even when a disconnect is triggered by afk timeout

### LINKED ISSUE #203: [BUG] - If 'num reconnect attempts' setting is > 0, reconnects will be attempted even when a disconnect is triggered by timeout.
**UE Version:**
5.2

**Problem component**
Frontend

**Description**
If `num reconnect attempts` setting is greater than 0, then reconnects will be attempted even when a disconnect is triggered by timeout.

**Expected behavior**
Reconnects should not be attempted when the stream is closed due to timeout.

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
This PR addresses https://github.com/EpicGames/PixelStreamingInfrastructure/issues/203.

## Solution
When programmatically closing the websocket connection, we set the `shouldReconnect` bool to false, indicating that the webRtcPlayerController should no attempt to auto reconnect.

## Documentation
No updates needed.

Testing completed by triggering a stream restart on the frontend and observed the auto reconnect functionality not being triggered.

Also triggered the reconnect flow and observed it still worked as expected.

### DIFF
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.test.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.test.ts
@@ -209,7 +209,7 @@ describe('PixelStreaming', () => {
         expect(webSocketSpyFunctions.constructorSpy).toHaveBeenCalledTimes(1);
         expect(webSocketSpyFunctions.closeSpy).not.toHaveBeenCalled();
 
-        pixelStreaming.disconnect();
+        pixelStreaming.webSocketController.close();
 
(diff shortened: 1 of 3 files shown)


mcottontensor (approved):

adrianZahra:
Looks good.

PR #232, opened 2023-05-17
Added "media-playout" to prevent spam in Aggregated Stats

chasse20:
## Relevant components:
- [x ] Frontend library

## Problem statement:
Many browser versions don't yet support the "media-playout" stats type which causes flooding spam in the console, making it impossible to debug. This is a simple fix to stop that.

See: [media-playout](https://developer.mozilla.org/en-US/docs/Web/API/RTCStatsReport)

## Solution
Default cases it out like similar other stats. Also it appears Typescript hasn't updated their definition of RTCStatsType enum to include "media-playout" yet. That's annoying.


lukehb (approved):
Approved. Sorry for the delay.

PR #246, opened 2023-06-06
Fixing hovering mouse mode set in URL being overridden on refresh.

### LINKED ISSUE #222: Control Scheme `HoveringMouse` doesn't activate via URL and resets on refresh
**UE Version:**
UE 5.2

**Frontend Version:**
0.6.2

**Problem component**
Frontend

**Description**
`HoveringMouse` is unable to be set via URL. If set via the UI, it will reset to `LockedMouse` when the stream is refreshed or restarted.

**Steps to Reproduce:**
1. Launch any Pixel Streaming enabled application with local streaming arguments;
2. Launch the Signalling Server;
3. Connect to the application locally via the browser;
4. Add `?HoveringMouse=true` to the URL;
5. Click into the stream and observe the mouse getting locked instead of hovering.
6. Escape to regain control of the mouse, remove the URL switch and change _Control Scheme_ to _Hovering Mouse_;
7. Observe the mouse switching to hovering mode correctly and the UI add `?HoveringMouse=true` to the URL;
8. Refresh the page or restart the stream;
9. Observe _Control Scheme_ reset to _Locked Mouse_, while the URL retains `?HoveringMouse=true`.

**Expected behavior**
1. Adding `?HoveringMouse=true` to the URL switches _Control Scheme_ to _Hovering Mouse_.
2. _Control Scheme_ retains its value if the stream is refreshed or restarted in the same window.

**Environment**
 - Chrome 112.0.5615.138 
 - Windows 10

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
When setting the mouse mode via the URL the frontend will ignore it when setting initial settings even if the initial settings don't include a value for it. See #222

## Solution
We only set the mouse mode if the setting exists in the initial settings.

Check that the UI setting is changing the mouse mode. Then check that the URL setting is setting the mouse mode, even on refresh. Then check that an application can set the mouse mode with the initial settings. This can override the URL.

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -1857,10 +1857,10 @@ export class WebRtcPlayerController {
                 parsedInitialSettings.PixelStreaming;
         }
 
-        if (parsedInitialSettings.ConfigOptions) {
+        if (parsedInitialSettings.ConfigOptions && !!parsedInitialSettings.ConfigOptions.DefaultToHover) {
             this.config.setFlagEnabled(
                 Flags.HoveringMouseMode,
-                !!parsedInitialSettings.ConfigOptions.DefaultToHover
+                true
             );
         }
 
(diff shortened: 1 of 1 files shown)


lukehb (approved):
lgtm

PR #247, opened 2023-06-06
Fixing viewport resizing not always working due to improperly calling timer.

### LINKED ISSUE #239: [BUG] - Frontend - Feature MatchViewportRes not quite working.
**UE Version: **
E.g. UE 5.1.1

**Frontend Version: **
0.5.1

**Problem component**
Frontend

**Description**
The Flag **MatchViewportRes** doesn't seems to alway do its job, in particular when click and dragging the side of the browser. 

This seems to be caused by the event filtering that is happening here 
https://github.com/EpicGames/PixelStreamingInfrastructure/blob/eea7839f513056ebbd0f98930eb8b8ca895d6bf1/Frontend/library/src/VideoPlayer/VideoPlayer.ts#L213 
(removing that filter fix the problem)

In particular the function referenced by setTimeout seems to never beeing called due to the timeout
https://github.com/EpicGames/PixelStreamingInfrastructure/blob/eea7839f513056ebbd0f98930eb8b8ca895d6bf1/Frontend/library/src/VideoPlayer/VideoPlayer.ts#L232 

Frontend dev is not my speciality so I am not sure why this is hapenning and how to fix it but I'm curious to understand what's going on here.

**Steps to Reproduce:**
1. Set the MatchViewportRes flag to 'true' in the react frontend implementation
2. Start react frontend
3. Resize the window by click and dragging the size of the browser. 
4. Observe the black bands around the video stream (ie the resolution of the stream doesn't match the resolution of the window.

 - OS: Windows 11 Pro
 - Chrome Version 113.0.5672.127

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Sometimes when resizing the viewport with the match viewport resolution option turned on, the application will not properly resize. See #239 

## Solution
When the resize event comes in too fast, the resize is offloaded to a timer but the timer was not properly calling the resize function. This just fixes the call. The times have also been slightly adjusted to make it feel a little more responsive.

### DIFF
--- a/Frontend/library/src/VideoPlayer/VideoPlayer.ts
+++ b/Frontend/library/src/VideoPlayer/VideoPlayer.ts
@@ -210,7 +210,7 @@ export class VideoPlayer {
         }
 
         const now = new Date().getTime();
-        if (now - this.lastTimeResized > 1000) {
+        if (now - this.lastTimeResized > 300) {
             const videoElementParent = this.getVideoParentElement();
             if (!videoElementParent) {
                 return;
@@ -230,8 +230,8 @@ export class VideoPlayer {
             );
             clearTimeout(this.resizeTimeoutHandle);
             this.resizeTimeoutHandle = window.setTimeout(
-                () => this.updateVideoStreamSize,
-                1000
+                () => this.updateVideoStreamSize(),
+                100
             );
         }
     }


Frontend/library/src/VideoPlayer/VideoPlayer.ts
```diff
@@ -230,8 +230,8 @@ export class VideoPlayer {
             );
             clearTimeout(this.resizeTimeoutHandle);
             this.resizeTimeoutHandle = window.setTimeout(
-                () => this.updateVideoStreamSize,
-                1000
+                () => this.updateVideoStreamSize(),
```
lukehb, line 233: Oh, that's good.

lukehb (approved):
Approved.

Issue #258, opened 2023-06-08
Full screen Issue in IOS safari


vijodvijay:
Both Issue are in IOS safari
UE 5.1.1

Pixel streaming plugin
if you switch the pixel streaming to full screen then it runs as video as video player and the inputs will not work anymore.

In UE 5.2
You can do full screen it doesn't work

lukehb:
CC @Belchy06 this sounds vaguely familiar to me (Matt will not be able to repro on his side)

lukehb:
Hey @vijodvijay - this is fixed in https://github.com/EpicGames/PixelStreamingInfrastructure/pull/266

PR #266, opened 2023-06-12
Re-enable iOS and iPadOS fullscreen

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
This issue addresses https://github.com/EpicGames/PixelStreamingInfrastructure/issues/258, which states full screen on iOS has been broken since moving to the typescript frontend.

This regression is caused by iOS only allowing video elements to be full screened, whereas iPad and all other devices allow arbitrary html elements to become full screen. [source](https://developer.apple.com/forums/thread/133248)

## Solution
This PR addresses the problem by first checking the browsers user agents and if the device is an iPod or iPhone, we will only full screen the video element, otherwise we will full screen the parent div which contains all of the interactable UI. 

NOTE: As previously described, iPhones and iPods can't full screen arbitrary elements and as such requesting full screen will result in the video being played through the native player and input being prevented.

Tested on an iPhone and iPad to confirm correct operation between devices.

### DIFF
--- a/Frontend/ui-library/src/Application/Application.ts
+++ b/Frontend/ui-library/src/Application/Application.ts
@@ -194,7 +194,9 @@ export class Application {
             ? new FullScreenIconExternal(this._options.fullScreenControlsConfig.customElement)
             // Or use the one created by the Controls initializer earlier
             : controls.fullscreenIcon;
-        if (fullScreenButton) fullScreenButton.fullscreenElement = this.rootElement;
+        if (fullScreenButton) {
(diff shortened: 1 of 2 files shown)


lukehb (approved):
Beautiful.

copostic:
Hi guys ! 
Is it planned to bring this PR to 5.1 ?

Belchy06:
Hey @copostic,

As mentioned in the PR, iOS will still use the native player regardless of whether you're using the typescript frontend or the js frontend in 5.1 and earlier.

That being said, I'll look at bringing the user agent check which will allow full screen on iPads into the older branches tomorrow.

PR #274, opened 2023-06-16
Fix iOS touch when settings panel is open

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Touch events weren't routed to panels once a video had started playing. This prevented modification of settings that we're not visible as you couldn't scroll to see them.

## Solution
Remove the passive aspect from touch handlers. This may have some negative impact on page performance but is necessary as the listeners we registers aren't passive.

Tested on mobile platforms to ensure fix as well as desktop platforms to ensure no regressions


lukehb (approved):
lgtm

gunsha:
removing the `passive: false` is throwing errors on touch events on chrome emulating mobile
SCR-20230704-mowt

Issue #281, opened 2023-06-27
[BUG] - When using OfferToReceive, if you refresh during a stream, a second offer is sent.


mcottontensor:
**UE Version: **
UE 5.3

**Frontend Version: **
UE5.3-0.0.3

**Problem component**
Frontend

**Description**
As per title. If you start viewing a stream using OfferToReceive and while viewing the stream you hit refresh on the browser, the player will go through the normal disconnection behaviour but it seems that immediately after another player connection is made and another offer is sent, followed immediately by another player disconnect.

**Steps to Reproduce:**
1. Start a Pixel Streaming application
2. Open the browser to the frontend
3. Open settings and set Browser Send Offer to true.
4. Start the stream.
5. Hit refresh on the browser.
6. Notice the logs in the signalling server showing, disconnect, connect, offer, disconnect.

**Expected behavior**
There should be no new connection from the player until they click to start the stream again.

**Platform (please complete the following information):**
OS: Windows 10 Pro
Browser + version: Firefox 114.0.2

PR #285, opened 2023-06-29
Fixing auto reconnect to not reconnect when the page is refreshed.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Described in #281
When the option "OfferToReceive" is set, if the user refreshes the page during a stream the frontend will immediately try to reconnect to the SS before immediately disconnecting again.

## Solution
The on close event now checks the code of the event to check whether the close event was caused by the local machine or not.


Belchy06 (approved):
LGTM

PR #303, opened 2023-07-18
Show player count in stats panel

### LINKED ISSUE #302: Is there an option to get current number of users in front-end UI?
Is there any option to get current number of users on an single stream instance.
To show in pixel streaming frontend UI. That's updated dynamically for all the users running the stream

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [X] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Player count information is simply logged to the developer console but is never shown in any UI, making it harder for non-technical users to view this information as reported in https://github.com/EpicGames/PixelStreamingInfrastructure/issues/302

## Solution
Add a new `playerCount` event that user's can listen to. By default, UI Library will listen for this event and update the value in the stats panel.

### DIFF
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.ts
@@ -25,7 +25,8 @@ import {
     WebRtcConnectingEvent,
     WebRtcDisconnectedEvent,
     WebRtcFailedEvent,
-    WebRtcSdpEvent
+    WebRtcSdpEvent,
+    PlayerCountEvent
 } from '../Util/EventEmitter';
 import { MessageOnScreenKeyboard } from '../WebSockets/MessageReceive';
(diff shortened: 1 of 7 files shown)


mcottontensor:
The player count currently just tracks the number of players that connect to the signalling server. With the ability to handle multiple streamers now this count may be misleading if there are multiple streamers connected.
If we're ok with that for now, I'm happy for this to go in.

Belchy06:
Hmm, good point.

I think with that fact there may need to be some more though put into this. 

That being said, I think we can get this landed and look to moving to players per stream as part of the great signalling server refactor.

Thoughts?

mcottontensor:
Yeah landing this can still be of use. We will just need an enhancement issue added to track the players per stream work.

mcottontensor (approved):

PR #311, opened 2023-07-18
Add: DataChannel Latency Test

gingernaz:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Measure data channel latency

## Solution
Introducing a new test that will show data channel latency values

## Documentation
To start the test:
- Open stats panel
- Hit `Run Test` button under `Data Channel Latency Test`
- Wait for the results

Compatible with the latest main


lukehb:
Just confirming backwards compatibility is preserved? 

For example, are the following cases met:

1. For users on older versions of UE does this silently fail (if so, good!)? 
2. Likewise, for the UE side changes, if they are using the old frontend (without this change) does it still stream no problems (if so good!).

mcottontensor (approved):

gingernaz:
> Just confirming backwards compatibility is preserved?
> 
> For example, are the following cases met:
> 
> 1. For users on older versions of UE does this silently fail (if so, good!)?
> 2. Likewise, for the UE side changes, if they are using the old frontend (without this change) does it still stream no problems (if so good!).


Just double-checked this.
Users with older versions of UE will get `NaN` values with the test. Users of the older frontend are still able to stream with no problems.

gingernaz:
Can we please drop this PR in favour of https://github.com/EpicGames/PixelStreamingInfrastructure/pull/315

PR #316, opened 2023-07-19
Ensure touch is relative to absolute location of parent rect

### LINKED ISSUE #287: [BUG] - Touch input location is shifted when embeded in web page
**Problem component**
 Frontend - Touch input
"@epicgames-ps/lib-pixelstreamingfrontend-ue5.2": "^0.5.1",
 "@epicgames-ps/lib-pixelstreamingfrontend-ui-ue5.2": "^0.4.1",

**Description**
When using a tablet with touch input and the pixel streaming player is embeded in a webpage like the showcase.html in streaming server public the location of the touch is shifted.
For example if you have a sidebar in the html page and the unreal "window" is onle part of the width the touch position is shifted for the space of the sidebar.

**Steps to Reproduce:**
Using a windows tablet with Chrome, Firefox or Edge an the showcase.html in streaming server public. Or a webpage were the unreal "window" covers not the complete browser.
For me it occured with locked mouse, hovering mouse, FakeMouseWithTouches false or true.

Thank you for your great work.

StomyPX:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
#287 

## Solution
Fixes #287 by making touch coordinates be relative to the viewport's absolute bounding rect rather than its own stated offset

Uses a different class member to retrieve the video element's offset on screen, but the underlying logic is identical.

Tested by streaming a plain 3rd person project to an Android phone that previously reproduced the problem stated above.

### DIFF
--- a/Frontend/library/src/Inputs/TouchController.ts
+++ b/Frontend/library/src/Inputs/TouchController.ts
@@ -152,16 +152,15 @@ export class TouchController implements ITouchController {
         if (!this.videoElementProvider.isVideoReady()) {
             return;
         }
-        const videoElementParent =
-            this.videoElementProvider.getVideoParentElement();
+        const offset = this.videoElementProvider.getVideoParentElement().getBoundingClientRect();
         const toStreamerHandlers =
             this.toStreamerMessagesProvider.toStreamerHandlers;
 
         for (let t = 0; t < touches.length; t++) {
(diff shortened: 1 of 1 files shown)


Belchy06 (approved):
LGTM

PR #336, opened 2023-07-27
Add: Protocol structures can now contain strings

### LINKED ISSUE #331: Expose `registerMessageHandler` for datachannel messages
**Component your feature request relates to**
- [ ] Frontend library

**Problem your feature solves? Please describe.**
Currently a user of frontend library is unable to register handlers for custom datachannel messages or override the stock options.

**Describe the solution you'd like**
A user of the library should be to add a custom message handler using public methods.

**Additional considerations**

- At first glance this seems like just exposing a method to be public, however there are additional side effects of doing so that must also be handled by whomever implements this ticket;
- We'll need to somehow expose the default senders we have (one for structured messages like `KeyDown` and the other for non-structured like `Command`),  so users can override these.

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Fixes https://github.com/EpicGames/PixelStreamingInfrastructure/issues/331. 

Additionally, while fixing the issue I noted that there was two separate code paths when sending messages to the streamer, one for descriptors (`Command`, `UIInteraction`) and one for structured message (`Key(Up | Down | Press)`, `Mouse(Enter | Leave | Move | Down | Up | Double)`, `Gamepad(Analog | ButtonPressed | ButtonReleased)`).

## Solution
- Add public method `registerMessageHandler` on the `PixelStreaming` class
- Add public getter for the `toStreamerHandlers` on the `PixelStreaming` class
- Remove `sendDescriptorController`. The functionality in `sendDescriptor` has been rolled into `sendMessageToStreamer`
- Remove `TwoWayMap`. Both times it was used could be replaced with simple `Map` data types

Using the two new methods on the `PixelStreaming` class, users can add a new input message type and call it whenever they want or they can add a custom handler for sending/receiving their new message. eg 
#### Triggering a new message to be sent to the streamer
```typescript
const config = new Config({ useUrlParams: true });
const stream = new PixelStreaming(config);

// Add MyMessage to the list of handlers.
stream.registerMessageHandler('MyMessage', MessageDirection.ToStreamer)

// Trigger MyMessage to be sent to the streamer with the specified data
stream.toStreamerHandlers.get('MyMessage')(['Hello', 123])
```

#### Triggering a new message to be sent to the streamer with a custom sender
```typescript
const config = new Config({ useUrlParams: true });
const stream = new PixelStreaming(config);

// Add MyMessage to the list of handlers.
stream.registerMessageHandler('MyMessage', MessageDirection.ToStreamer, (data: Array<number | string>) => {
        // TODO: Implement custom sender
        // eg data will contain ['Hello', 123]
        console.log(data.map((element: string | number) => typeof element)) // string, number
})

// Trigger MyMessage to be sent to the streamer with the specified data
stream.toStreamerHandlers.get('MyMessage')(['Hello', 123])
```

#### Handling a new message from the streamer
```typescript
const config = new Config({ useUrlParams: true });
const stream = new PixelStreaming(config);

// Add MyMessage to the list of handlers.
stream.registerMessageHandler('MyMessage', MessageDirection.FromStreamer, (data: ArrayBuffer) => {
        // TODO: Implement custom receiver
})
```

PixelStreaming tests passed

| Unreal Engine Branch | Pixel Streaming Infra Branch | Result |
| ------------------------ | ------------------------------- | ------- |
| UE5/Main                    | Belchy06/master                    | ✔️ |
| UE5/Main                    | EpicGames/master                 | ⚠️ Warnings logged about malformed protocol as new message layout doesn't contain `byteLength` but this is to be expected and all functionality is retained |
| UE5/Release-5.3          | Belchy06/master                    | ✔️ |
| UE5/Release-5.3          | EpicGames/master                 | ✔️ |
| N/A                                        | Belchy06/master                | Tests:       36 passed, 36 total  |


Frontend/library/src/Inputs/KeyboardController.ts
```diff
@@ -232,10 +232,7 @@ export class KeyboardController {
         Logger.Log(Logger.GetStackTrace(), `key up ${keyCode}`, 6);
         const toStreamerHandlers =
             this.toStreamerMessagesProvider.toStreamerHandlers;
-        toStreamerHandlers.get('KeyUp')([
-            keyCode,
-            keyboardEvent.repeat ? 1 : 0
-        ]);
+        toStreamerHandlers.get('KeyUp')([ keyCode ]);
```
lukehb, line 235: `repeat` is removed because this was caught as extraneous data that was added to the handler but not actually even being sent or used on the UE side!

Frontend/library/src/PixelStreaming/PixelStreaming.ts
```diff
@@ -732,4 +733,33 @@ export class PixelStreaming {
...
+        if(direction === MessageDirection.FromStreamer && typeof handler === 'undefined') {
+            Logger.Warning(Logger.GetStackTrace(), `Unable to register an undefined handler for ${name}`)
+            return;
+        }
+
+        if(direction === MessageDirection.ToStreamer && typeof handler === 'undefined') {
+            this._webRtcController.streamMessageController.registerMessageHandler(
+                direction,
+                name,
+                (data: Array<number>) =>
```
lukehb: only `Array<number>` here?

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1711,8 +1717,14 @@ export class WebRtcPlayerController {
...
         Logger.Log(Logger.GetStackTrace(), `WebRTC FPS=${fps}`, 6);
         if (fps != null) {
-            this.sendDescriptorController.emitCommand({'WebRTC.Fps': fps});
-            this.sendDescriptorController.emitCommand({'WebRTC.MaxFps': fps}); /* TODO: Remove when UE 4.27 unsupported. */
+            this.streamMessageController.toStreamerHandlers.get(
+                'Command'
+            )([JSON.stringify({'WebRTC.Fps': fps})]);
+
+            /* TODO: Remove when UE 4.27 unsupported. */
+            this.streamMessageController.toStreamerHandlers.get(
```
lukehb, line 1725: May as well remove this here then

lukehb (changes requested):
Left some minor comments.

Frontend/library/src/PixelStreaming/PixelStreaming.ts
Belchy06: Should have been `Array<number | string>`. Fixed.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
Belchy06: I'm going to leave this in as removing it would break functionality for users who use a single frontend for multiple version of UE

lukehb (approved):

PR #354, opened 2023-09-04
Consume the context menu event instead of sending a mouse up

StomyPX:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Right-Click dragging is presently impossible in hovering mouse mode as when the frontend receives a context menu event this gets instead sent as a right-mouse-up to Unreal Engine which is largely nonsensical.

## Solution
This PR simply consumes the context menu event and allows UE to decide for itself when to open a context menu which it can already do reliably. In exchange, right-click dragging now works correctly.

## Documentation
N/A

Simply launching an editor stream and testing out RMB-dragging in any viewport or node-graph shows the restored functionality.

### DIFF
--- a/Frontend/library/src/Inputs/HoveringMouseEvents.ts
+++ b/Frontend/library/src/Inputs/HoveringMouseEvents.ts
@@ -86,6 +86,7 @@ export class HoveringMouseEvents implements IMouseEvents {
         if (!this.mouseController.videoElementProvider.isVideoReady()) {
             return;
         }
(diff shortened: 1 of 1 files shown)


lukehb (approved):
Approved. Can confer will @Belchy06 on which PSInfra branches this is relevant to merge back to down to (UE 5.3 at least seems relevant).

PR #358, opened 2023-09-08
Auto connect on new streamer

mcottontensor:
## Relevant components:
- [x] Signalling server
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
If the front end tries to join a stream before the streamer is connected it requires a click to connect once the streamer has connected.

## Solution
This adds a new signalling message to the signalling server that broadcasts to all players that a new streamer has connected to the signalling server. The front end will respond to this if it's waiting, by auto joining the new stream.

None


Frontend/library/src/WebSockets/SignallingProtocol.ts
```diff
@@ -6,6 +6,7 @@ import {
     MessageRecvTypes,
     MessageConfig,
     MessageStreamerList,
+    MessageStreamerConnected,
```
lukehb: Hmmm, I would really prefer us to not add a new protocol message until we have a mechanism to transmit the signaling protocol and version.

If we absolutely can't achieve this another way then I guess by our own rules:

https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/SignallingWebServer/Docs/SignallingProtocol.md

This qualifies as a minor version bump in the signalling protocol (as it is independent and not a required message).

Could we also just sit in a timer and try again every few seconds if we couldn't connect and got back no streamers connected. I know it is more wasteful - but also no signaling change.

What do you think @mcottontensor?

lukehb (dismissed):

Frontend/library/src/WebSockets/SignallingProtocol.ts
mcottontensor: Fair. I can update it to just be a timer.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1399,6 +1406,15 @@ export class WebRtcPlayerController {
         }
     }
 
+    startAutoJoinTimer() {
+        clearTimeout(this.autoJoinTimer);
+        this.autoJoinTimer = setTimeout(() => this.tryAutoJoin(), 3000);
```
lukehb: We should expose this time interval as a setting too.

lukehb (changes requested):
Tiny change to expose interval as a setting.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: No problem.

Frontend/library/src/Config/Config.ts
```diff
@@ -565,6 +566,19 @@ export class Config {
                 useUrlParams
             )
         );
+
+        this.numericParameters.set(
+            NumericParameters.WaitRetryDelay,
+            new SettingNumber(
+                NumericParameters.WaitRetryDelay,
```
lukehb: Can this be called like `StreamerAutoJoinInterval` instead?

lukehb (approved):
Approved with suggestion to change most recent setting name.

lukehb (approved):

PR #369, opened 2023-09-14
Remove unit conversion for bitrate from URL. URL is already in kbps

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Setting the WebRTC bitrate in the URL would lead to the settings panel being incorrectly populated  on stream restart.

## Solution
The URL already specifies the bitrate in kbps so there is no need to do a unit conversion when populating settings panel.


lukehb (approved):
Approved

PR #389, opened 2023-10-17
Update SignallingWebServer platform scripts to support Mac

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [X] Platform scripts
- [ ] SFU

## Problem statement:
Although PixelStreaming doesn't currently support Mac as a platform, this doesn't mean that the infrastructure can't be run on Mac (x86_64 / Arm64) devices.

## Solution
This PR modifies the bash platform scripts to first check for the OS and architecture before grabbing the appropriate Node binaries. 

Additionally, I've forked CoTurn and created binaries for each Mac architecture. This platform scripts now grab the correct binary for the machine's architecture as well.

## Documentation
No documentation updates as we didn't explicitly state that we didn't support Mac.

### DIFF
--- a/SignallingWebServer/platform_scripts/bash/Start_TURNServer.sh
+++ b/SignallingWebServer/platform_scripts/bash/Start_TURNServer.sh
@@ -25,7 +25,15 @@ echo ""
 
 # Hmm, plain text
 realm="PixelStreaming"
-process="turnserver"
+process=""
+if [ "$(uname)" == "Darwin" ]; then
+	process="${BASH_LOCATION}/coturn/bin/turnserver"
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+	process="turnserver"
+else
+	echo 'Incorrect host OS for use with Start_TURNServer.sh'
+	exit -1
+fi
 arguments="-c turnserver.conf --allowed-peer-ip=$localip -p ${turnport} -r $realm -X $publicip -E $localip -L $localip --no-cli --no-tls --no-dtls --pidfile /var/run/turnserver.pid -f -a -v -u ${turnusername}:${turnpassword}"
 
 # Add arguments passed to script to arguments for executable
--- a/SignallingWebServer/platform_scripts/bash/Start_TURNServer.sh
+++ b/SignallingWebServer/platform_scripts/bash/Start_TURNServer.sh
@@ -28,11 +28,8 @@ realm="PixelStreaming"
 process=""
 if [ "$(uname)" == "Darwin" ]; then
 	process="${BASH_LOCATION}/coturn/bin/turnserver"
-elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
-	process="turnserver"
 else
-	echo 'Incorrect host OS for use with Start_TURNServer.sh'
-	exit -1
+	process="turnserver"
 fi
 arguments="-c turnserver.conf --allowed-peer-ip=$localip -p ${turnport} -r $realm -X $publicip -E $localip -L $localip --no-cli --no-tls --no-dtls --pidfile /var/run/turnserver.pid -f -a -v -u ${turnusername}:${turnpassword}"
 
--- a/SignallingWebServer/platform_scripts/bash/setup.sh
+++ b/SignallingWebServer/platform_scripts/bash/setup.sh
@@ -141,11 +141,8 @@ if [ "$(uname)" == "Darwin" ]; then
 		echo 'Incompatible architecture. Only x86_64 and ARM64 are supported'
 		exit -1
 	fi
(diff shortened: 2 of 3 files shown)


SignallingWebServer/platform_scripts/bash/Start_TURNServer.sh
```diff
@@ -25,7 +25,15 @@ echo ""
 
 # Hmm, plain text
 realm="PixelStreaming"
-process="turnserver"
+process=""
+if [ "$(uname)" == "Darwin" ]; then
+	process="${BASH_LOCATION}/coturn/bin/turnserver"
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
lukehb: I would prefer just if darwin, then an else that just assumes we are on Linux and attempts to the Linux codepath as opposed to failing with an incorrect OS error (I just think this could have edge cases that would be problematic).

SignallingWebServer/platform_scripts/bash/setup.sh
```diff
@@ -128,10 +129,28 @@ node_version=""
...
+	arch=$(uname -m)
+	if [[ $arch == x86_64* ]]; then
+		node_url="https://nodejs.org/dist/$NODE_VERSION/node-$NODE_VERSION-darwin-x64.tar.gz"
+	elif  [[ $arch == arm* ]]; then
+	    node_url="https://nodejs.org/dist/$NODE_VERSION/node-$NODE_VERSION-darwin-arm64.tar.gz"
+	else
+		echo 'Incompatible architecture. Only x86_64 and ARM64 are supported'
+		exit -1
+	fi
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
lukehb: Same here, I would prefer to just defualt assume yes we are on Linux if we are not on Darwin.

SignallingWebServer/platform_scripts/bash/setup.sh
```diff
@@ -144,20 +163,38 @@ setup_frontend
...
+			coturn_url="https://github.com/Belchy06/coturn/releases/download/v4.6.2-mac-x84_64/turnserver.zip"
+		elif  [[ $arch == arm* ]]; then
+	    	coturn_url="https://github.com/Belchy06/coturn/releases/download/v4.6.2-mac-arm64/turnserver.zip"
+		fi
+		curl -L -o ./turnserver.zip "$coturn_url"
+		mkdir "${BASH_LOCATION}/coturn" 
+		tar -xf turnserver.zip -C "${BASH_LOCATION}/coturn"
+		rm turnserver.zip
+	fi
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
```
lukehb: Same again here.

lukehb (commented):
Left some comments about default assuming we are on Linux if we are not on Darwin.

lukehb (changes requested):
Requesting we default assume Linux if not on Darwin.

SignallingWebServer/platform_scripts/bash/setup.sh
Belchy06: Done!

SignallingWebServer/platform_scripts/bash/setup.sh
Belchy06: Done!

SignallingWebServer/platform_scripts/bash/setup.sh
```diff
@@ -144,20 +160,38 @@ setup_frontend
...
+if [ "$(uname)" == "Darwin" ]; then
+	if [ -d "${BASH_LOCATION}/coturn" ]; then
+		echo 'CoTURN directory found...skipping install.'
 	else
-		if [ `id -u` -eq 0 ]; then
-			check_and_install "coturn" "$coturn_version" "1" "apt-get install -y coturn"
+		echo 'CoTURN directory not found...beginning CoTURN download for Mac.'
+		coturn_url=""
+		if [[ $arch == x86_64* ]]; then
+			coturn_url="https://github.com/Belchy06/coturn/releases/download/v4.6.2-mac-x84_64/turnserver.zip"
```
lukehb: Let's get these into releases on the PSInfra repo instead.

May as well fix up the .bat version where it is grabbing from Matt's repo too.

lukehb (changes requested):
Let's grab the coturn bins from this repo instead of your personal repo (same with the Windows one grabbing from Matt's repo).

SignallingWebServer/platform_scripts/bash/setup.sh
Belchy06: Done!

lukehb (approved):
lgtm

PR #392, opened 2023-10-18
[UI-library] Prevent errors on disabled setting or stats panel.

timbotimbo:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
You can disable the settings and stats panel when using the ui-library.

```js
const application = new Application({
	stream,
	settingsPanelConfig: {
		isEnabled: false,
		visibilityButtonConfig: { creationMode: UIElementCreationMode.Disable }
	} ,
	statsPanelConfig: {
		isEnabled: false,
		visibilityButtonConfig: { creationMode: UIElementCreationMode.Disable }
	} 
});
```

However doing so might give you various errors because the `settingsPanel` or `statsPanel`  variable will be undefined.

- If you only disable the settings panel, the stats button will break.
The console shows an error as it tries `this.settingsPanel.hide();`.  

onclick

- If you only disable the stats panel, you get the reverse with the settings button breaking.
- However the stats panel is also called repeatedly in other functions like
`onPlayerCount`, `onVideoInitialized` and `onStatsReceived`.
Although these don't seem to break anything visible to the user, they throw hundreds of console errors during a short stream.

stats-error

## Solution
Adding some `?` operators on usage of `this.settingsPanel` and `this.statsPanel` to avoid throwing errors.

## Documentation
N / A

Some additional null checks shouldn't break any compatibility or features.

The stats and settings button don't break anymore in manual testing.

## Backporting
These errors show up starting with version 5.2 (which I use), but I'm making this pull request on master.
But because of code changes I'm not sure if a backport from master will fix all occurences in 5.2 and 5.3.

Please let me know If I need to make separate pull requests for other versions.

### DIFF
--- a/Frontend/ui-library/src/Application/Application.ts
+++ b/Frontend/ui-library/src/Application/Application.ts
@@ -487,7 +487,7 @@ export class Application {
      * Shows or hides the settings panel if clicked
      */
     settingsClicked() {
-        this.statsPanel.hide();
+        this.statsPanel?.hide();
         this.settingsPanel.toggleVisibility();
     }
 
@@ -495,7 +495,7 @@ export class Application {
      * Shows or hides the stats panel if clicked
      */
     statsClicked() {
-        this.settingsPanel.hide();
+        this.settingsPanel?.hide();
         this.statsPanel.toggleVisibility();
     }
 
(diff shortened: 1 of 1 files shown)


Belchy06 (approved):
LGTM!

lukehb (approved):
Yep, that's a good idea. Code looks good to me. Lgtm.

PR #400, opened 2023-10-19
Fix errors when setting a null preferred codec

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
If a user attempts to connect to a pixel stream but doesn't receive any video, the frontend will try to update the preferred codec dropdown with a null codec.

image

## Solution
We'll simply check that the codec is not null before we try to set the dropdown value.


lukehb (approved):
lgtm!

PR #402, opened 2023-10-20
Update SFU platform scripts to support Mac

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [X] Platform scripts
- [X] SFU

## Problem statement:
SFU doesn't currently support Mac due to using prebuilt binaries of Mediasoup. Additionally, the platform scripts for the SFU don't download Mac specific Node binaries.

## Solution
To address the SFU no supporting Mac, we can just use the normal npm package which works fine.

To fix the setup scripts, I've copied the same logic that's currently in the SignallingWebServer platform scripts.

Tested SFU on Mac on both the host and in a container.

### DIFF
--- a/SFU/package.json
+++ b/SFU/package.json
@@ -5,20 +5,19 @@
   "private": true,
   "scripts": {
     "start-local": "run-script-os --",
-      "start-local:windows": ".\\platform_scripts\\cmd\\run.bat",
-      "start-local:default": "./platform_scripts/bash/run_local.sh",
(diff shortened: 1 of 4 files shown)


lukehb:
@Belchy06 What was the status of this one? Did you confirm SFU working on Windows, Linux, and Mac?

Belchy06:
Cannot confirm and is proving tricker than I thought. I'll close this and come back with another one.

PR #403, opened 2023-10-23
Allowing the Matchmaker to run on linux.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [x] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The matchmaker script was failing to find the custom install of node.

## Solution
Like the other scripts, this change will update the PATH to include the custom install path.


lukehb (approved):
lgtm

Issue #350, opened 2023-08-17
[BUG] - SDP/ICE Failure when connecting during Endpoint ID Change


StomyPX:
**UE Version: **
UE 5.3

**Frontend Version: **
UE5.3

**Problem component**
Signalling Server/Frontend

**Description**
If a player attempts to subscribe to a streamer immediately after they've connected, the SDP/ICE negotiation may lock up requiring a refresh

**Steps to Reproduce:**
1. Launch signalling server
2. Launch packaged UE project
3. Open tab to PixelStreaming URL.
4. Count to 3
5. Click to begin streaming

The timing has to be just right (roughly 2-3 seconds) but if you can get the player to connect immediately after the streamer connects, you can catch it at the point where it still has `__LEGACY__` as its endpoint ID. This causes SDP/ICE negotiation to be initiated with the wrong endpoint ID and requires refreshing the page to connect again properly

**Expected behavior**
Ideally the signalling server could communicate the id change to the frontend and complete or restart the negotiation process without the user having to know. Or the endpoint ID could be passed as part of initially connecting the streamer to the signalling server and so not require any handling at all.

**Platform (please complete the following information):**
 - OS: Manjaro Linux
 - Browser: Google Chrome 115.0.5790.170

PR #411, opened 2023-10-25
Fixes for Streamers changing IDs when connecting & Updates to SFU behaviour relating to multi streamer changes.

mcottontensor:
## Relevant components:
- [x] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [x] SFU

## Problem statement:
Streamer IDs changing:
Streamers need to identify themselves so when they first connect they are assigned a temporary ID. If a player attempts to connect while this temporary name is assigned it can lead to stuck behaviour.

SFUs with multi streamer behaviour:
When the multi streamer change went in the SFU behaviour was kind of hacked into the flow. Only one SFU could be connected at all and it was assumed that there was always only one streamer.

Frontend build scripts smashing windows PATH:
When running the frontend build script, depending on the environment setup you can end up with an incorrect PATH variable. This is due to the script trying to set PATH using the registry variables, but if you manually add or modify PATH in your session before running the script, you will lose them.

## Solution
Streamer IDs changing:
Now when streamers connect, they are not "published" in the streamer list until they ID themselves. If they do not ID themselves within a time, they are assigned an ID. This prevents peers connecting to streamers during their ID negotiation.

SFUs with multi streamer behaviour:
There are several new configuration options with the SFU. The SFU ID, which is used to identify the SFU as a streamer on the signalling server. This ID will show up in the streamers list when the SFU is streaming. And the subscribe streamer ID which tells the SFU what streamer ID we are looking to forward. If it cannot find the specified streamer it will request the streamer list on a timeout that is specified in the config until it finds the streamer it's looking for.
Additionally the SFU can now start/stop streaming which will control when the SFU shows up in the streamer list on the signalling server. The SFU will start streaming when it connects to the streamer it's configured to forward, and stop streaming when it loses this connection.

Frontend build scripts smashing windows PATH:
Changed the setting via registry to a simple set OLDPATH and then set PATH to that temp variable.

## Documentation
Documentation has been added to the SFU configuration file. Documentation for the streamer ID fix is not needed.

I have manually run through various scenarios of connecting and disconnecting different components in different orders and switching between SFU and DefaultStreamers. We will need to do a usual QA pass over this change.


mcottontensor:
Fixes #350

SignallingWebServer/cirrus.js
```diff
@@ -348,20 +429,40 @@ class Player {
...
+		}
+		return this.sfuStreamerComponent;
+	}
+};
 
-function getSFU() {
-	return players.get(SFUPlayerId);
+let streamers = new Map();		// streamerId <-> streamer
+let players = new Map(); 		// playerId <-> player/peer/viewer
+const LegacyStreamerPrefix = "__LEGACY_STREAMER__"; // old streamers that dont know how to ID will be assigned this id prefix.
```
lukehb, line 452: Is changing this string problematic @Belchy06 ?

SignallingWebServer/cirrus.js
```diff
@@ -401,46 +502,112 @@ function getPlayerIdFromMessage(msg) {
...
 
+function getUniqueLegacyId() {
+	for (let i = 0; i < 99; ++i) {
+		const testId = LegacyStreamerPrefix + i;
+		if (!streamers.has(testId)) {
+			return testId;
+		}
+	}
+	return ""; // no available id
+}
```
lukehb, line 511: Legacy id creation change to be prefix + number, @Belchy06 is this also okay for backward compat?

SignallingWebServer/cirrus.js
```diff
@@ -401,46 +502,112 @@ function getPlayerIdFromMessage(msg) {
...
+		}
+		const numPostfix = Number(postfix);
+		if (numPostfix > maxPostfix) {
+			maxPostfix = numPostfix
+		}
+	}
+	if (maxPostfix >= 0) {
+		return id + (maxPostfix + 1);
+	}
+	return id;
```
lukehb, line 561: A bit curious what is happening here.

Is this correct: we do a check to see if the base id of an existing streamer matches the new id, then if they are matching, then increment the numeric part of the id, is that right?

Also is the base id an ip address now? 

So my summary here is I don't like the regex, it seems a bit fragile. Imagine if we change how we do the id in the future. Also regex is a sort of expensive operation to do in a linear search, granted it probably isn't called that often and the size of streamers should be small, but still it seems like not the right solution given our goal or unique ids here.

Can we store the id as a struct so we have the addr and increment stored as fields on the struct instead of encoding them both in a single string?

SignallingWebServer/cirrus.js
```diff
@@ -401,46 +502,112 @@ function getPlayerIdFromMessage(msg) {
...
 }
 
+function getUniqueLegacyId() {
+	for (let i = 0; i < 99; ++i) {
+		const testId = LegacyStreamerPrefix + i;
+		if (!streamers.has(testId)) {
+			return testId;
+		}
+	}
+	return ""; // no available id
```
lukehb: Instead of for-loop here can we just store a global legacyId increment. This for loop gets slower each unique streamer we add. I get the size of the loop is small, but still, I think legacyId increment lets us have unlimited legacy streamer ids and is O(1) to generate a new id.

SignallingWebServer/cirrus.js
```diff
@@ -401,46 +502,112 @@ function getPlayerIdFromMessage(msg) {
...
+					break;
+				}
+			}
+		}
+		if (available) {
+			return testId;
+		}
+	}
+	return ""; // no available id
+}
```
lukehb, line 518: Pretty much same comment as loop for generating legacy streamer id, can we just have a global for SFU id generator.

SignallingWebServer/cirrus.js
```diff
@@ -495,7 +662,8 @@ streamerServer.on('connection', function (ws, req) {
 	console.logColor(logging.Green, `Streamer connected: ${req.connection.remoteAddress}`);
 	sendStreamerConnectedToMatchmaker();
 
-	let streamer = { ws: ws };
+	const temporaryId = req.connection.remoteAddress;
+	let streamer = new Streamer(temporaryId, ws, StreamerType.Regular);
```
lukehb, line 653: So to be clear, streamer ids will be now set to their remote address (+ potentially some id appended to the back). Is there a separator between the remote address and the appended id?

SignallingWebServer/cirrus.js
```diff
@@ -535,69 +704,124 @@ streamerServer.on('connection', function (ws, req) {
...
+	console.log("disconnecting SFU from streamer");
+	disconnectAllPlayers(sfuPlayer.id);
+	onStreamerDisconnected(sfuPlayer.getSFUStreamerComponent());
+	sfuPlayer.unsubscribe();
+	sfuPlayer.ws.close(4000, "SFU Disconnected");
+	players.delete(sfuPlayer.id);
+	streamers.delete(sfuPlayer.id);
 }
 
+sfuMessageHandlers.set('listStreamers', onPlayerMessageListStreamers);
```
lukehb, line 786: Shouldn't `onPlayerMessageListStreamers` have some check to see if streamer id `commited` (just for safety as the code changes in the future)?

SignallingWebServer/platform_scripts/cmd/setup_frontend.bat
```diff
@@ -43,10 +43,10 @@
...
   set "NodeDir=%CD%\SignallingWebServer\platform_scripts\cmd\node"
 
-  @Rem Prepend NodeDir to PATH temporarily using a custom tool called SetEnv
-  call SignallingWebServer\platform_scripts\cmd\setenv\SetEnv.exe -uap PATH %%%%"%NodeDir%"
-  @Rem Refresh the cmd session with new PATH
-  call %~dp0\refreshenv.cmd
+  @rem Save the old path variable
+  set OLDPATH=%PATH%
+  @Rem Prepend NodeDir to PATH temporarily
+  set PATH=%PATH%;%NodeDir%
```
lukehb, line 49: Will want Michael to test this works on AWS

lukehb (commented):
Left some comments and questions

lukehb:
Given there is some changes to fairly load bearing code I am going to ask @Belchy06 to review also.

And once reviews have been done I'm gonna ask @MWillWallT to give this a QA pass:

- Using all the different infra components and configs on AWS - Linux and Windows
- Testing whether previous version of UE 5.0 - 5.3 still connect to this signalling server, SFU, and matchmaker.

SignallingWebServer/cirrus.js
Belchy06: Should be fine. 

We just assign them an ID so regardless of what they're assigned, the current code paths will be fine.

SignallingWebServer/cirrus.js
Belchy06: Same as above, should be fine.

SignallingWebServer/cirrus.js
mcottontensor: This is to just prevent 2 streamers from being the same ID. So if DefaultStreamer already exists and someone tries to register as DefaultStreamer, they will be assigned DefaultStreamer1 and the next DefaultStreamer2.
The id is only set to the address of the connecting streamer while we're waiting for it to register itself. It should never actually be used in anything other than logging.

SignallingWebServer/cirrus.js
mcottontensor: No. This is just temporary and allows us to get a better idea of what's happening if anything is logged while this streamer is negotiating its id. It should never actually be used for anything but logging.

SignallingWebServer/cirrus.js
mcottontensor: It could just for sanity but the `streamers` list should not contain anything non committed. `committed` is referring to the act of adding the streamer to that list.

SignallingWebServer/cirrus.js
lukehb: Okay understood. I still think the code/intent is clearer if we store id like

{
prefix: String,
increment: Number,
toString: ()->String
}

Not sure how much other stuff that messes up though. I guess it turns it into a pretty big refactor then?

Does the regex break if I set my streamer id to an integer say: `137`

SignallingWebServer/cirrus.js
lukehb: There is no UE side code that assumes `_LEGACY_` ?

lukehb: I can see `Engine\Plugins\Media\PixelStreaming\Source\PixelStreamingServers\Private\SignallingServer.cpp` does use that string. But I don't think that matters here, as that is simply the UE side signalling server.

Frontend/library/src/Config/Config.ts
```diff
@@ -315,17 +314,6 @@ export class Config {
...
-        this.flags.set(
-            Flags.PreferSFU,
-            new SettingFlag(
-                Flags.PreferSFU,
-                'Prefer SFU',
-                'Try to connect to the SFU instead of P2P.',
-                false,
-                useUrlParams
-            )
-        );
```
lukehb, line 327: Is the idea here that now that we support multiple streamers, they just select the SFU just like any other streamer? If so, agreed.

Frontend/ui-library/src/Config/ConfigUI.ts
```diff
@@ -174,10 +174,6 @@ export class ConfigUI {
             psSettingsSection,
             this.flagsUi.get(Flags.StartVideoMuted)
         );
-        this.addSettingFlag(
-            psSettingsSection,
-            this.flagsUi.get(Flags.PreferSFU)
-        );
```
lukehb, line 180: Removing this setting means you should also update this page: 

https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/Frontend/Docs/Settings%20Panel.md

lukehb (changes requested):
Few more questions and spotted a doc change that needs to go in too.

SignallingWebServer/cirrus.js
mcottontensor: I think that's needlessly complex because we just want to come up with a unique name in the case of name clashes. Once the name is generated there's no benefit to breaking it down since it's all just reported to the user via the streamer list.
Addressed the issue where the streamer id is totally numeric now.

Frontend/library/src/Config/Config.ts
mcottontensor: Yes. There is also room to improve it slightly to report the "type" of the streamer in the streamer list in the case where someone wants to "prefer" an sfu over other available streamers, programmatically.

Frontend/ui-library/src/Config/ConfigUI.ts
mcottontensor: Updated.

lukehb (approved):

lukehb:
@mcottontensor Michael has given it the green light for QA, I have approved. You may want to consider which branches benefit from the backport using the autobackport tags.

PR #419, opened 2023-11-01
Rewriting a bunch of reconnect and disconnect handling.

### LINKED ISSUE #401: [BUG] - MaxPlayerCount can result in a reconnection loop.
**UE Version:** 
Tested with 5.2, but not relevant.

**Frontend Version:** 
5.2, 5.3, 5.4
Master branch  right now (current commit https://github.com/EpicGames/PixelStreamingInfrastructure/commit/2a21ee656618d778e86f088255048e742fa5d6c0)

(I tested 5.2 and master, so I assume 5.3 and 5.4 are the same)

**Problem component**
Frontend library.

**Description**
The frontend will endlessly try to reconnect if the SignallingServer has reached the player limit and there is no Matchmaker configured.

Using the setup:
- 1 unreal instance
- 1 SignallingServer without Matchmaker.
- MaxPlayerCount >0 
- Multiple browser windows.

**Steps to Reproduce:**
1. Download the repo and run the setup scripts for the SignallingServer.
2. Run any unreal project with Pixelstreaming.
3. Edit `SignallingWebServer/config.json` to set `"MaxPlayerCount": 1`
4. Start the server using `Start_SignallingServer.ps1`.
5. Visit the ip of the SignallingServer to reach the player.html example and start the connection.  (this one works as expected)
6. Open a second browser window doing the same.
7. The UI in the second window will now be stuck on **Auto connecting now**.

The `WaitForStreamer` and `MaxReconnectAttempts` config options don't seem to make a difference either.

**Expected behavior**
The frontend should show a connection error or detect that it can stop trying to reconnect.

**Screenshots**
The browser console keeps repeating these logs.
browser

The SignallingServer keeps outputting this error.
server

 - OS: Windows 10 Pro.
 - FireFox & Chrome, any version

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
In addressing #401 I noticed there was a lot of clumsiness when dealing with disconnections and reconnections. There were several situations that would lead to reconnection loops, or incorrect status reporting etc.

## Solution
I have cleaned up and streamlined a bunch of the logic around when to connect and which streamers to auto subscribe to. I have also made changes to how the reporting is done on the overlay.
Fixes #401

A lot of manual tests and the usual npm tests. Will need further QA.


mcottontensor:
Fixes #401

Frontend/library/src/Util/EventEmitter.ts
```diff
@@ -347,7 +347,9 @@ export class StreamerListMessageEvent extends Event {
         /** Streamer list message containing an array of streamer ids */
         messageStreamerList: MessageStreamerList;
         /** Auto-selected streamer from the list, or null if unable to auto-select and user should be prompted to select */
-        autoSelectedStreamerId: string | null;
+        autoSelectedStreamerId: string;
```
lukehb, line 350: If null type is no longer supported then the comment above should change too?

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -104,16 +104,13 @@ export class WebRtcPlayerController {
...
     shouldReconnect: boolean;
     isReconnecting: boolean;
     reconnectAttempt: number;
+    disconnectMessageOverride: string;
     subscribedStream: string | null;
     signallingUrlBuilder: () => string;
-
-    // if you override the disconnection message by calling the interface method setDisconnectMessageOverride
-    // it will use this property to store the override message string
-    disconnectMessageOverride: string;
```
lukehb, line 115: lol, what a pointless comment it had before

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -223,28 +209,41 @@ export class WebRtcPlayerController {
...
+            this.pixelStreaming._onDisconnect(disconnectMessage, !willTryReconnect && !this.isReconnecting);
+
             this.afkController.stopAfkWarningTimer();
 
             // stop sending stats on interval if we have closed our connection
             if (this.statsTimerHandle && this.statsTimerHandle !== undefined) {
                 window.clearInterval(this.statsTimerHandle);
             }
 
+            this.setVideoEncoderAvgQP(0);
```
lukehb, line 233: Maybe add comment here, this may be unintuitive that (I assume) the reason this is done is to reset the stream quality icon.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -947,9 +946,9 @@ export class WebRtcPlayerController {
     }
 
     /**
-     * Restart the stream automatically without refreshing the page
+     * Attempt a reconnection to the signalling server
      */
-    restartStreamAutomatically() {
```
lukehb, line 952: Thank you. Thank you. Thank you. Much better name.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -959,33 +958,16 @@ export class WebRtcPlayerController {
...
-            );
+        // if the connection is open, first close it. wait some time and try again.
+        this.isReconnecting = true;
+        if (this.webSocketController.webSocket && this.webSocketController.webSocket.readyState != WebSocket.CLOSED) {
+            this.closeSignalingServer(`${message} Restarting stream...`);
+            setTimeout(() => {
+                this.tryReconnect(message);
+            }, 3000);
+        } else {
             this.pixelStreaming._onWebRtcAutoConnect();
```
lukehb, line 970: This one is the other one with a bad name imo.

Frontend/library/src/PixelStreaming/PixelStreaming.ts
```diff
@@ -353,7 +352,7 @@ export class PixelStreaming {
      */
     public reconnect() {
         this._eventEmitter.dispatchEvent(new StreamReconnectEvent());
-        this._webRtcController.restartStreamAutomatically();
```
lukehb, line 356: Nice.

Frontend/library/src/PixelStreaming/PixelStreaming.ts
```diff
@@ -353,7 +352,7 @@ export class PixelStreaming {
...
+        this._webRtcController.tryReconnect("");
```
lukehb: Maybe text in this case could be "Reconnecting..." ?

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -104,16 +104,13 @@ export class WebRtcPlayerController {
     preferredCodec: string;
     peerConfig: RTCConfiguration;
     videoAvgQp: number;
+    locallyClosed: boolean;
     shouldReconnect: boolean;
     isReconnecting: boolean;
     reconnectAttempt: number;
+    disconnectMessageOverride: string;
```
lukehb: Do we actually need this property at all?

Frontend/ui-library/src/Application/Application.ts
```diff
@@ -667,18 +667,41 @@ export class Application {
...
+            } else if (messageStreamingList.ids.length == 0) {
+                if (isReconnecting) {
+                    message = `Waiting for a streamer to become available.`;
+                    allowRestart = false;
+                } else {
+                    message = `No streamers available. Click to try again.`;
+                    allowRestart = true;
+                }
+            } else {
+                message = `Multiple streamers available. Select from the settings menu.`;
```
lukehb: Perhaps
 
 > Multiple streamers available. Select one from the settings menu.
 
 > Multiple streamers available. Select a streamer from the settings menu.

lukehb (changes requested):
Left a few minor comments.

lukehb:
@MWillWallT This PR changes some loading bearing code around reconnections and disconnections.

Can I get you to test the following:

- Reconnection/disconnect working as expected with one streamer
- Reconnection/disconnect working as expected with one streamer and multiple peers
- AFK working as expected with one streamer and multiple peers
- Killing the streamer midstream then putting it back up auto reconnects fine
- Killing the streamer midstream then waiting for reconnects to exhaust, putting the streamer back up then clicking connect works fine
- Change the reconnect time interval and attempts works as expected
- Connecting/disconnecting/reconnecting works fine when there are multiple streamers
- Matchmaker with multiple UEs works fine
- Anything else you think that would make sense for reconnection flows.
- Tested on Chrome, Firefox, Safari, Android, and iOS.

Also keep in mind some of the messages displayed on the frontend change in this PR. Might be worth seeing if all the common ones are worded correctly and make sense for the scenario they show up in.

Frontend/library/src/Util/EventEmitter.ts
mcottontensor: string type allows null by default.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: Honestly I'm as clueless as you. I just copied this from existing behaviour.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: tryReconnect? or _onWebRtcAutoConnect?

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
lukehb: I am pretty sure it is what I said, maybe just add a comment above it saying

> This is done to reset the stream quality icon.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
lukehb: `_onWebRtcAutoConnect()`

lukehb: Don't have to do anything about it in this PR, but if the opportunity comes up that one is also in my sights.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
mcottontensor: Yeah we still set it internally and use it later. I've updated the name to be less verbose, however.

MWillWallT:
I've gone through and tested as many re-connection configs as I can, as well as all the suggested you mentioned @lukehb 
All good on this front

lukehb (approved):

lukehb:
QA has passed, you can consider backporting to 5.2, 5.3 too with this one I think.

PR #420, opened 2023-11-01
Exposing the pixelstreaming interface to the browser.

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [x] Frontend source
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Pixelstreaming is inaccessible to the browser and hidden away in the onload function. This prevents any ability for the browser to access the pixelstreaming interface for things like automated testing.

## Solution
A global is now declared that can be accessed when using the browser context. This allows automation scripts to query information or even control the public interface of pixelstreaming.


lukehb (approved):
All good!

Issue #275, opened 2023-06-20
Cannot set min/max bitrate at stream start time from frontend


DenisTensorWorks:
**UE Version: **
UE 5.2

**Problem component**
Frontend

**Description**
Min/MaxBitrate unable to be set from frontend at stream start time, neither it can be configured during stream running.
It does, however, work if the stream is re-started.

Reason being, UE sends its initial settings message which contains the information about min/max bitrate, which overrides what’s set on the frontend before the frontend has a chance to send its URL values.

For context, if the limitations are added to application launch args instead, i.e. `-PixelStreamingEncoderMaxBitrate=5000`, they work as expected.

Suggested fix is to send this info as a config message.

mcottontensor:
This is a tricky one since the config would need to be sent before the stream starts but we have no connection to the endpoint until the webrtc connection is negotiated and streaming.
Perhaps there is something we can put in the SDP.

lukehb:
I do think when we loop round to this, we should consider putting the metadata into the SDP - that actually seems pretty reasonable considering what we are trying to achieve here.

PR #425, opened 2023-11-09
Bitrate negotiation on stream startup

mcottontensor:
- Added min/max bitrate requests for offer/answer messages
- Fixing bitrate values in the frontend units and parsing.

## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [x] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Setting the bitrate values before starting the stream did not work.

## Solution
Added min/max bitrate values to the answer signalling message so the streamer can extract them and set them when needed.

Will probably need a QA pass.

### DIFF
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -1551,7 +1551,11 @@ export class WebRtcPlayerController {
             'Sending the offer to the Server',
             6
         );
-        this.webSocketController.sendWebRtcOffer(offer);
+        
+        const minBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMinBitrate);
+        const maxBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMaxBitrate);
+
+        this.webSocketController.sendWebRtcOffer(offer, minBitrate, maxBitrate);
     }
 
     /**
@@ -1564,7 +1568,11 @@ export class WebRtcPlayerController {
             'Sending the answer to the Server',
             6
         );
-        this.webSocketController.sendWebRtcAnswer(answer);
+
+        const minBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMinBitrate);
+        const maxBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMaxBitrate);
+
+        this.webSocketController.sendWebRtcAnswer(answer, minBitrate, maxBitrate);
 
         if (this.isUsingSFU) {
             this.webSocketController.sendWebRtcDatachannelRequest();
--- a/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
+++ b/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
@@ -1,6 +1,6 @@
 // Copyright Epic Games, Inc. All Rights Reserved.
 
-import { WebSocketController } from '../WebSockets/WebSocketController';
+import { WebSocketController, ExtraOfferParameters, ExtraAnswerParameters } from '../WebSockets/WebSocketController';
 import { StreamController } from '../VideoPlayer/StreamController';
 import {
     MessageAnswer,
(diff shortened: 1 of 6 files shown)


Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1551,7 +1551,11 @@ export class WebRtcPlayerController {
             'Sending the offer to the Server',
             6
         );
-        this.webSocketController.sendWebRtcOffer(offer);
+        
+        const minBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMinBitrate);
+        const maxBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMaxBitrate);
+
+        this.webSocketController.sendWebRtcOffer(offer, minBitrate, maxBitrate);
```
lukehb: Rather than appending more parameters here (as I am certain we will want more in the future) can we instead pass an offerOptions, offerParams, offerConfig, or something like this as the additional parameter to this function/

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1564,7 +1568,11 @@ export class WebRtcPlayerController {
             'Sending the answer to the Server',
             6
         );
-        this.webSocketController.sendWebRtcAnswer(answer);
+
+        const minBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMinBitrate);
+        const maxBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMaxBitrate);
+
+        this.webSocketController.sendWebRtcAnswer(answer, minBitrate, maxBitrate);
```
lukehb: Same comment as above about an options param.

Frontend/library/src/WebSockets/MessageSend.ts
```diff
@@ -24,17 +24,26 @@ export class MessageSend implements Send {
     type: string;
     peerConnectionOptions: object;
 
+    /**
+     * A filter for controlling what parameters to actually send.
+     * Good for excluding default values or hidden internals.
+     * Return undefined to exclude the property completely.
+     */
+    sendFilter(key: string, value: any) {
```
lukehb, line 37: Good idea. Can comment have an example of how it should be used.

lukehb (changes requested):
Left some comments about organising params into an struct/config object.

Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts
```diff
@@ -1551,7 +1551,11 @@ export class WebRtcPlayerController {
             'Sending the offer to the Server',
             6
         );
-        this.webSocketController.sendWebRtcOffer(offer);
+        
+        const minBitrate = 1000 * this.config.getNumericSettingValue(NumericParameters.WebRTCMinBitrate);
```
lukehb: Consider naming as minBitrateBps, although technicaly bitrate implies bps, the word bitrate is also used for kilobits and megabits quite often so I think the extra clarity here helps.

lukehb (changes requested):
Minor changes to naming/structuring really.

Frontend/library/src/WebSockets/WebSocketController.ts
```diff
@@ -159,13 +169,13 @@ export class WebSocketController {
         this.webSocket.send(payload.payload());
     }
 
-    sendWebRtcOffer(offer: RTCSessionDescriptionInit, minBitrate: number, maxBitrate: number) {
-        const payload = new MessageSend.MessageWebRTCOffer(offer, minBitrate, maxBitrate);
+    sendWebRtcOffer(offer: RTCSessionDescriptionInit, extraParams: ExtraOfferParameters) {
+        const payload = new MessageSend.MessageWebRTCOffer(offer, extraParams.minBitrateBps, extraParams.maxBitrateBps);
```
lukehb: Can the constructor of MessageSend.MessageWebRTCOffer and MessageSend.MessageWebRTCAnswer take the ExtraParameters object for the same reason (so constructor doesn't grow overlong as we add more).

lukehb (changes requested):
Minor comment about using ExtraParams object even more.

Frontend/library/src/WebSockets/MessageSend.ts
```diff
@@ -134,7 +149,7 @@ export class MessageWebRTCAnswer extends MessageSend {
     /**
      * @param answer - Generated Web RTC Offer
      */
-    constructor(answer: RTCSessionDescriptionInit, minBitrate: number, maxBitrate: number) {
+    constructor(answer: RTCSessionDescriptionInit, extraParams: ExtraOfferParameters) {
```
lukehb: Offer params passed to answer?

lukehb (changes requested):
Minor correct to type used I think

Frontend/library/src/WebSockets/MessageSend.ts
lukehb: Also comment:

> @ param answer - Generated Web RTC Offer

Seems wrong in the same way, talking about offer in answer constructor.

lukehb (approved):
lgtm!

DenisTensorWorks:
This PR closes this issue: https://github.com/EpicGames/PixelStreamingInfrastructure/issues/275

PR #428, opened 2023-11-16
Update platform scripts so that the BASH_LOCATION works on Mac as well

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [X] Platform scripts
- [ ] SFU

## Problem statement:
The logic for detecting the script location is only suitable for bash. This PR makes the logic work on both bash and zsh.

Tested script execution on Mac and Linux


lukehb (approved):
If this has passed QA lgtm!

PR #431, opened 2023-11-20
Dockerfile for SFU

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [X] SFU

## Problem statement:
There was an old Dockerfile in the repo but it was old and in a slightly different location to other components of infra.
Additionally there were some errors in the bash platform_scripts that prevented it from building the frontend.

## Solution
Updates the Dockerfile and moves it into the expected location.
Also updated the platform_scripts for bash to properly pass arguments around.

## Documentation
Will add docker instructions to the SFU documentation in another PR when the previous docs PR is in.


lukehb (commented):
Looks good, one thing is, let's get a github workflow as part of this PR to auto build this image and push it to the official containers, much like this one: https://github.com/EpicGames/PixelStreamingInfrastructure/blob/master/.github/workflows/container-images.yml

lukehb:
Actually the docs on the front page should change if we do ship a container on the SFU too.

lukehb (approved):
lgtm!

PR #432, opened 2023-11-20
Fixing left behind streamer IDs when they're late to ID themselves.

mcottontensor:
## Relevant components:
- [X] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
If a streamer was slow to start up, it can be late to ID itself and the signalling server will assign it a legacy id. When the streamer finally replies it will get assigned a new ID but the old ID will be left behind leading to bad streamer lists with non existent streamer IDs.

## Solution
When a streamer gets assigned an ID the server will check that any old ID is removed from the streamer list.


SignallingWebServer/cirrus.js
```diff
@@ -524,6 +524,10 @@ function sanitizeStreamerId(id) {
 }
 
 function registerStreamer(id, streamer) {
+	// remove any existing streamer id
+	if (!!streamer.id) {
+		streamers.delete(streamer.id);
+	}
```
lukehb, line 540: CC @Belchy06 any problems you forsee by doing this?

Belchy06: Might just need to check if the reconnect flow still works.

I'd think that if a user connected when the streamer was assigned the legacy ID, it wouldn't be able to reconnect once the new streamer had ID'd itself.

mcottontensor: I've updated the behaviour by handling any previously subscribed players and sending them an event. The frontend now also will respond to the id change event by updating its subscribed streamer name.

Belchy06: Changes look good.

One last thing, though. With the streamer id changed event, do you know if the dropdown in the settings panel updates as well?

Belchy06: Hmm, looks like we'll need to add:

```typescript
this.config.setOptionSettingValue(
    OptionParameters.StreamerId,
    newID
);
```

to `handleStreamerIDChangedMessage` or bind to the new `StreamerIDChangedMessageEvent` where we set the option.

mcottontensor: Added this functionality. It wasn't that simple for various reasons but it now appears to update the settings dropdown properly.

SignallingWebServer/Docs/SignallingProtocol.md
```diff
@@ -229,6 +230,14 @@ end
 | Param Name | Type | Description |
 |-|-|-|
 
+### streamerIDChanged<a name="signalling-streameridchanged"></a>
+
+> Message is used to communicate to [Player](#term-player)s that the [Streamer](#term-streamer) it is currently subscribed to is changing it's ID. This allows Players to keep track of it's currently subscribed Streamer and allow auto reconnects to the correct Streamer. This happens if a Streamer sends an [endpointID](#streamer-endpointid) message after it already has an ID assigned. (Can happen if it is late to respond to the [identify](#signalling-identify) message and is auto assigned a legacy ID.)
```
Belchy06: Just fixing a few `it's` => `its`

> Message is used to communicate to [Player](#term-player)s that the [Streamer](#term-streamer) it is currently subscribed to is changing its ID. This allows Players to keep track of its currently subscribed Streamer and allow auto reconnects to the correct Streamer. This happens if a Streamer sends an [endpointID](#streamer-endpointid) message after it already has an ID assigned. (Can happen if it is late to respond to the [identify](#signalling-identify) message and is auto assigned a legacy ID.)

PR #438, opened 2023-11-22
Fixing handling of StreamerId in the initial config settings.

### LINKED ISSUE #436: [BUG] - Frontend ignores the StreamerId set in the config.
**UE Version:**
Tested with 5.3, likely not relevant

**Frontend Version:**
Tested with 5.4, might also apply to 5.3 and 5.2.

**Problem component**
Frontend

**Description**
Setting the `StreamerId` option in the config doesn't work as expected.

With urlParams enabled, adding `?StreamerId=SFU` to the url works as expected, it auto selects the streamer with that id.
Now i'm trying to do the same but without using the url.

**Steps to Reproduce:**
1. Use typescript implementation project.
2. Make sure 2 streamers are available, in my case by starting the SFU.
3. Edit `player.ts`:
Disable urlParams and set the streamerId in the config definition.
```ts
// set streamerid either in initialSettings or with setSettings
// enter your applicable streamerId for SFU
const config = new Config({ useUrlParams: false, initialSettings:{'StreamerId': 'SFU'}});
config.setSettings({'StreamerId': 'SFU'});

stream = new PixelStreaming(config);
``` 
4. View the UI in the browser and see that the StreamerId from the config is not being used.

**Expected behavior**
The application auto selects the streamer set with `StreamerId` (if it is available), just like when it is set in the url.

**Platform:**
 - OS: Windows 10
 - Browser: Firefox 119.0.1, Chrome 119.0.6045.160

**Additional context**
I'm experimenting with 1 streamer and 1 SFU to see if this helps with slow connections.
Previously the `PreferSFU` option auto connected to the SFU, but with the removal of that option, the 5.4 frontend will show an option to select one of multiple streamers.

I'm trying to work around this by setting 'SFU' as the streamerId on startup.

**More context**
Checking where the StreamerId is autoselected, it appears to only check the url for params.
https://github.com/EpicGames/PixelStreamingInfrastructure/blob/12a08e1624d07547dcf85454b152872cffac48c3/Frontend/library/src/WebRtcPlayer/WebRtcPlayerController.ts#L1339-L1347

mcottontensor:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
As described in #436 setting the StreamerId in the initial config does not work. This is because if we try to set an option selection that isn't already in the list the selection is ignored, and additionally when we get the streamer list, this option list is thrown out and reset with the streamer ids.

## Solution
- The setOptionSettingValue now adds the selected value if it doesn't already exist in the list. Without this the selection does not happen.
- handleStreamerListMessage now checks the existing selection and uses it as the default wantedStreamerId. It also now correctly checks whether we're allowing URL params before checking them for the StreamerId.

Fixes #436

### DIFF
--- a/Frontend/library/src/Config/Config.ts
+++ b/Frontend/library/src/Config/Config.ts
@@ -742,7 +742,13 @@ export class Config {
                 `Cannot set text setting called ${id} - it does not exist in the Config.enumParameters map.`
             );
         } else {
-            this.optionParameters.get(id).selected = settingValue;
+            const optionSetting = this.optionParameters.get(id);
+            const existingOptions = optionSetting.options;
+            if (!existingOptions.includes(settingValue)) {
+                existingOptions.push(settingValue);
+                optionSetting.options = existingOptions;
+            }
+            optionSetting.selected = settingValue;
         }
     }
 
--- a/Frontend/library/src/Config/SettingOption.ts
+++ b/Frontend/library/src/Config/SettingOption.ts
@@ -24,7 +24,7 @@ export class SettingOption<
 		// eslint-disable-next-line @typescript-eslint/no-empty-function
 		defaultOnChangeListener: (changedValue: unknown, setting: SettingBase) => void = () => { /* Do nothing, to be overridden. */ }
     ) {
-        super(id, label, description, [defaultTextValue, defaultTextValue], defaultOnChangeListener);
+        super(id, label, description, [defaultTextValue], defaultOnChangeListener);
 
         this.options = options;
         const urlParams = new URLSearchParams(window.location.search);
--- a/Frontend/library/src/PixelStreaming/PixelStreaming.ts
+++ b/Frontend/library/src/PixelStreaming/PixelStreaming.ts
@@ -566,6 +566,10 @@ export class PixelStreaming {
 
         const useUrlParams = this.config.useUrlParams;
         const urlParams = new URLSearchParams(window.location.search);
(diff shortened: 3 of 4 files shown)


Frontend/library/src/Config/Config.ts
```diff
@@ -742,7 +742,13 @@ export class Config {
...
         } else {
-            this.optionParameters.get(id).selected = settingValue;
+            const optionSetting = this.optionParameters.get(id);
+            const existingOptions = optionSetting.options;
+            if (!existingOptions.includes(settingValue)) {
+                existingOptions.push(settingValue);
+                optionSetting.options = existingOptions;
+            }
+            optionSetting.selected = settingValue;
         }
```
lukehb, line 752: Few questions:

1. Which bit of code actually calls `setOptionSettingValue`?

2. Is it worth having a `setOrAddOptionSettingValue` - my concern is adding if missing is bit unexpected based on the name `setOptionSettingValue`?

Wdyt?

mcottontensor: It's only used in two places. One for the preferred codec and one for the streamer ids. My logic for adding it in the existing behaviour is that if you're trying to set the selected value of an option selector, it doesn't really make sense for it to fail silently and not do anything.

lukehb (approved):
lgtm

Issue #194, opened 2023-04-09
How to input Chinese in a input widget?


zixixr:
Hi there,

I created a chatbox in my multi-player PS experience, currently I could only input English on both PC and mobile phones. I can't even copy-paste Chinese in the chatbox. How can I solve it? Thanks.

DenisTensorWorks:
Hi zixixr,

Thank you for submitting this issue. Our team is currently looking into it, it looks like frontend only supports mapping English characters at the moment. We will keep you posted with updates.

zixixr:
Hi @DenisTensorWorks 

Thanks. And even for English characters, there are issues with inputting on mobile devices. You have to use the original input method for ios, and clicking a small type text icon to input.

jimcccc:
Ok

chase03:
https://blog.csdn.net/qq_40544338/article/details/116719305     same issue on Unreal Engine 5.0.3, may this help you, try set localization "zh" package. 
 I have tried this, failed to input Chinese character and only can texting English at this moment according Epic fellow

zhangshengsheng:
That's the problem with me, too

zhangshengsheng:
。。。。。。

DenisTensorWorks:
Reopening this issue as it is still in our backlog to fix. Thank you everyone for your patience with this.

AJI1026:
Same issue

PR #442, opened 2023-11-28
Support composition for mobile keyboards

gunsha:
## Relevant components:
- [ ] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
Currently when using mobile keyboards using the suggestion bar or asian keyboards (chinese/japanese) is not supported as this uses the composition event on the keyboard

## Solution
added composition event handler on mobile devices 

working example of the changes implemented
https://arcanemirage.com/project/394?token=hLVbruBPrnR8

posible fix for #194

### DIFF
--- a/Frontend/library/src/Inputs/KeyboardController.ts
+++ b/Frontend/library/src/Inputs/KeyboardController.ts
@@ -1,12 +1,12 @@
 // Copyright Epic Games, Inc. All Rights Reserved.
 
-import { SpecialKeyCodes } from './SpecialKeyCodes';
-import { Logger } from '../Logger/Logger';
-import { ActiveKeys } from './InputClassesFactory';
-import { StreamMessageController } from '../UeInstanceMessage/StreamMessageController';
 import { Config, Flags } from '../Config/Config';
+import { Logger } from '../Logger/Logger';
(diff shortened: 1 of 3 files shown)


lnsantos (approved):

lukehb:
Since this PR has been opened this commit was landed: https://github.com/EpicGames/PixelStreamingInfrastructure/commit/8b838f48fee68410c9334dc603ff93c54dfa8f1e

Is there other changes in this PR beyond support for composition events that we should consider? If so, please update the PR to the latest changes and update the description with any other issues it is addressing. However, I believe we support composition on mobile with the above linked commit?

lukehb:
Bump @gunsha for you comments on the above. ^

gunsha:
@lukehb can't find time to do a real test with the repo, but looking at the changes it's kind of the same I implemented, the only difference I see is that the composition and keypress events are always registered, i saw some input duplication when having both at the same time, so i'm only initializing composition on mobile and keypress on desktop

lukehb:
@gunsha Let us know if you do see any duplication using what is in master currently. I am going to close for now as it is likely a new PR would be required if there is any bugfix for duplicated input.

PR #472, opened 2024-01-15
Export NO_SUDO so it's available for use in launched processes

Belchy06:
## Relevant components:
- [ ] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [X] Platform scripts
- [ ] SFU

## Problem statement:
When launching the platform script on Mac, it always tries to launch the npm process with sudo, even if you pass `--nosudo` on the command line.

## Solution
By exporting the `NO_SUDO` variable, the npm process use it's value when determining whether to launch cirrus with `sudo` or not.


lukehb (approved):
lgtm

PR #473, opened 2024-01-23
Common library and Signalling Tester

mcottontensor:
## Relevant components:
- [x] Common library [NEW]
- [x] Signalling server
- [x] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [x] Platform scripts
- [ ] SFU

## Problem statement:
There are a bunch of code in the front end library that can be useful for other users. Without a common library the other components need to copy/paste or implement their own etc.

## Solution
A few common implementations are extracted into a standalone Common library. This allows implementers to link with this library and write their own frontend application.
Additionally a protobuf description for signalling messages has been added. This isn't being used by any of the existing functionality yet, but should be possible to bring this in in the future.
A new tool has also been added under SS_Test. This is the signalling protocol tester. This is a small test application which can test the validity of the signalling protocol between signalling server and streamers and peers.
This tool comes with a github action that will run on changes to several components and fail if anything in the protocol changes to break existing behaviour.

## Documentation

The following classes and types and now in Pixel Streaming common...
- Logger
- WebSocketTransport (Previously WebSocketController)
- MessageReceive
- MessageSend
- SignallingProtocol

 This change means that imports will have to change. For example, importing Logger would change from:
`import { Logger } from '../Logger/Logger';`
to
`import { Logger } from '@epicgames-ps/lib-pixelstreamingcommon-ue5.4';`

`WebSocketController` has now mostly moved into `SignallingProtocol`. Most of the functionality has been moved into this new class but it is not a 1 to 1 change so some changes might be required when upgrading.

Binding to the protocol message events in now done through the messageHandlers event emitter and its addListener method. Here is an example where we add a handler for when we receive the config message over the protocol:
```
messageHandlers.addListener(MessageReceive.MessageRecvTypes.CONFIG, (msg: MessageReceive.MessageRecv) => {
 // do something with config message
});
```


.github/workflows/publish-common-library-to-npm.yml
```diff
@@ -0,0 +1,23 @@
+name: Publish common lib
+on:
+  push:
+    workflow_dispatch:
+    branches: ['UE5.4']
```
lukehb: Based on our discussion, this trigger needs to be on the UE5.5 branch now.

.github/workflows/run-library-unit-tests.yml
```diff
@@ -27,10 +27,22 @@ jobs:
         node-version: '16.x'
```
lukehb: This should be node 18.17.0 to match the other instance of us using NodeJS.

.github/workflows/run-signalling-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    steps:
+    - name: "Checkout source code"
+      uses: actions/checkout@v3
+
+    - name: "Docker compose"
+      uses: isbang/compose-action@v1.5.1
+      with:
+        compose-file: "SS_Test/docker-compose.yml"
+        up-flags: "--build --abort-on-container-exit --exit-code-from tester"
```
lukehb, line 31: This looks great!

Common/.gitignore
```diff
@@ -0,0 +1,5 @@
+dist/
+docs/
```
lukehb: Why ignore `docs/` ?

Common/README.md
```diff
@@ -0,0 +1,21 @@
...
+An interface to a transport protocol that is in charge of sending and receiving signalling messages. Users can make use of supplied transport protocols or implement their own transport protocol via extending this interface. They can then pass the transport into the constructor of `SignallingProtocol` which is explained below.
+
+### WebSocketTransport
+An `ITransport` implementation that sends signalling messages over a websocket. This is currently the only transport in use by the official pixel streaming frontend.
+
+### SignallingProtocol
+This is the object where the user should send/receive messages. Currently there are specific functions for specific messages sent, but in the future this should change to be more generic. You can specify your own transport protocol by passing an implementation of `ITransport` to this class, allowing you to send and receive messages through any protocol you wish.
+
+### Adding it to your project
+`npm i @epicgames-ps/lib-pixelstreamingcommon-ue5.4`
```
lukehb: -ue5.5 suffix instead.

Common/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@epicgames-ps/lib-pixelstreamingcommon-ue5.4",
```
lukehb: -ue5.5

Common/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "version": "0.0.1",
+    "description": "Common utilities library for Unreal Engine 5.4 Pixel Streaming",
```
lukehb: 5.5

Common/protobuf/signalling_messages.proto
```diff
@@ -0,0 +1,132 @@
...
+  string sdp = 2;
+  optional string playerId = 3;
+}
+
+message peerConnectionOptions {
+}
+
+message config {
+  string type = 1;
+  peerConnectionOptions peerConnectionOptions = 2;
```
lukehb, line 19: What do you think about adding an optional signalling protocol version field here? I think this will be helpful going forward.

Common/protobuf/signalling_messages.proto
```diff
@@ -0,0 +1,132 @@
...
+
+message disconnectPlayer {
+  string type = 1;
+  string playerId = 2;
+  optional string reason = 3;
+}
+
+message endpointId {
+  string type = 1;
+  string id = 2;
```
lukehb, line 35: I am also considering an optional version field here where the streamer can specify its version, wdyt?

Common/src/Logger/Logger.ts
```diff
@@ -74,7 +71,11 @@ export class Logger {
      * @param message - the message to be logged
      */
     static Warning(stack: string, message: string) {
-        const returnString = `Level: Warning\nCaller: ${stack}\nMsg: ${message}`;
-        console.warn(returnString);
+        this.CommonLog("Warning", stack, message);
+    }
+
+    static CommonLog(level: string, stack: string, message: string) {
+        //console.log(`Level: ${level}\nMsg: ${message}\nCaller: ${stack}`);
```
lukehb: I think it is totally reasonable to not log the stack trace for `Info` and `Warning` messages, but I think for `Error` messages then we should. Can you make this modification?

Common/src/Transport/ITransport.ts
```diff
@@ -0,0 +1,12 @@
+import { MessageSend } from '../WebSockets/MessageSend';
+import { MessageRecv } from '../WebSockets/MessageReceive';
+
+export interface ITransport {
```
lukehb, line 10: You wrote a good description of this interface in the readme.md, can you also duplicate it here?

Common/src/WebSockets/MessageSend.ts
```diff
@@ -52,13 +52,6 @@ export class MessageSend implements Send {
     }
 }
 
-export interface Send {
-    /**
-     * Turns the wrapper into a JSON String
-     * @returns - JSON String of the Message to send
-     */
-    payload: () => string;
```
lukehb, line 60: The payload() method is removed from the public interface, we should document this in our PR notes and instruct users why/what to do different if they were using that method before.

Common/src/WebSockets/SignallingProtocol.ts
```diff
@@ -0,0 +1,108 @@
...
+import * as MessageReceive from './MessageReceive';
+import * as MessageSend from './MessageSend';
+
+/**
+ * Signalling protocol for handling messages from the signalling server.
+ */
+export class SignallingProtocol {
+    private transport: ITransport;
+    private messageHandlers: Map<string, (msg: MessageReceive.MessageRecv) => void>;
+    transportEvents: EventTarget;
```
lukehb: Do each of the transportEvents we report, e.g. `open`, `error`, `close` etc. have a unit test? If not, I think we could potentially skip seeing as we have the SS_Tester test anyway. Your thoughts?

Frontend/library/package.json
```diff
@@ -8,7 +8,7 @@
     "sideEffects": false,
     "scripts": {
         "compile": "tsc --build --clean && tsc",
-        "build": "npx webpack --config webpack.prod.js",
+        "build": "cd ../../Common && npm run build && cd ../Frontend/library && npm link ../../Common && npx webpack --config webpack.prod.js",
         "build-dev": "npx webpack --config webpack.dev.js",
```
lukehb: I think `build-dev` needs an update here too.

Frontend/library/src/PixelStreaming/PixelStreaming.ts
```diff
@@ -805,8 +805,8 @@ export class PixelStreaming {
      * Public getter for the websocket controller. Access to this property allows you to send
      * custom websocket messages.
      */
-    public get webSocketController() {
-        return this._webRtcController.webSocketController;
```
lukehb, line 809: Removing this public method from the API, I think the idea here is `signallingProtocol()` is a direct replacement for how the controller was used. If so we should mention this in our PR notes.

Frontend/library/src/WebSockets/WebSocketController.ts
```diff
@@ -1,260 +0,0 @@
...
-    close() {
-        this.webSocket?.close();
-    }
-
-    /**
-     * The Message Contains the payload of the peer connection options used for the RTC Peer hand shake
-     * @param messageConfig - Config Message received from he signaling server
-     */
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
-    onConfig(messageConfig: MessageReceive.MessageConfig) {}
```
lukehb, line 211: We remove this type from the public API in favour of signalling protocol, how would a downstream user now do custom handling of the `onConfig` message and so on.

I think this can be okay if we have events that are emitted for onConfig and others.

To my reading it seems these public functions were removed and the protocol messages are all bound within `WebRTCPlayerController`, which is fine, but we should also exposed events for those who need to do custom handling on these events.

lukehb: I think the ` addMessageHandler(messageId: string, messageHandler: (msg: MessageReceive.MessageRecv) => void) ` in the signallingprotocol class is not enough to fulfil this presently either because it can only take a single event bound per message type. One way to allow custom handling would be to allow addMessageHandler to, under to hood, support having multiple observers bound per message type.

lukehb: And ofc, this would then have to be documented in the PR notes.

SS_Test/src/index.ts
```diff
@@ -0,0 +1,129 @@
...
+  const entries1 = Object.entries(obj1);
+  const entries2 = Object.entries(obj2);
+
+  return (
+    entries1.length === entries2.length &&
+    entries1.every(([key, value]) => obj2.hasOwnProperty(key) && obj2[key] === value)
+  );
+}
+
+async function main(): Promise<void> {
```
lukehb, line 33: Small comment here explaining in general terms what `main()` procedure will attempt to do.

lukehb (changes requested):
Left some comments to address before landing, hmu for any as needed. Good work on this PR!

lukehb:
I guess our main readme.md should also include an explanation for what is `common` and `SS_Tester`

Common/.gitignore
mcottontensor: Hmm this file was copied directly from the frontend folder where its ignored there too.

Common/src/Logger/Logger.ts
mcottontensor: Upon looking at it again... Why do we pass in the stacktrace at the callsite rather than just in the method itself? It would mean that the stack trace ends with the Logger function but the very next level will have the one we're looking for. I think its a decent tradeoff to having large verbose logging calls all over the codebase.

Common/src/WebSockets/MessageSend.ts
mcottontensor: It was moved into MessageSend with a common implementation. I believe it can still be overridden, no?

Frontend/library/src/WebSockets/WebSocketController.ts
mcottontensor: So I've made a change to the handlers so it now uses an EventEmitter that's exposed in SignallingProtocol so we can use that functionality to add/remove however many listeners we want.
I also moved the transport events to an EventEmitter too mostly to be consistent but also because it's just a nicer interface.

Common/src/WebSockets/SignallingProtocol.ts
mcottontensor: I'm not sure what we would test? That they're fired when the websocket opens/closes/errors? That's working as it is and I don't see that part of the code changing a lot or even being very complex. I feel like adding unit tests to it would just add complexity.
The SS_Tester doesn't really touch that stuff at the moment.

Common/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "cspell": "^4.1.0",
+        "eslint": "^8.11.0",
+        "jest": "^27.5.1",
+        "jest-environment-jsdom": "27.5.1",
+        "prettier": "2.8.3",
+        "ts-jest": "27.1.5",
+        "ts-loader": "^9.4.2",
+        "typedoc": "^0.23.24",
+        "typescript": "^4.9.4",
+        "webpack": "^5.75.0",
```
lukehb: I don't think common uses webpack, jest, webxr. These can be removed.

lukehb: Additionally, we can probably add a `lint` under the `scripts` section to match frontend.

Common/src/WebSockets/SignallingProtocol.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Listen on this emitter for transport events, open, close, error.
+     */
+    transportEvents: EventEmitter;
+
+    /**
+     * Listen on this emitter for messages. Message type is the name of the event to listen for.
+     */
+    messageHandlers: EventEmitter;
```
lukehb, line 23: Nice

lukehb (changes requested):
Minor comments mostly about common's package.json

Common/src/WebSockets/SignallingProtocol.ts
lukehb: Fair, it is a bit like testing getter and setter at that point I suppose.

lukehb (approved):

Issue #10, opened 2022-07-11
Refactor SignallingWebServer to be consumable as an extensible libarary


mentalnote:
Similar to https://github.com/EpicGames/PixelStreamingInfrastructure/issues/9, we want to update the SignallingWebServer scripts so they can be interfaced with as a library.

This will more easily allow users to simply integrate the functionality they need from the library into their own infrastructure, along with the benefits described in the referred to issue above.

Names for this library up for discussion.

jimcccc:
Ok

DenisTensorWorks:
This issue has been moved to the new repository: https://github.com/EpicGamesExt/PixelStreamingInfrastructure/issues/1

PR #480, opened 2024-01-31
style unification

Senseme:
## Relevant components:
- [x] Signalling server

## Problem statement:
code style does not uniform

## Solution
unify the code style


lukehb:
We are actively in the process of completely rewriting the new signalling library as per https://github.com/EpicGames/PixelStreamingInfrastructure/issues/10 - the existing cirrus.js will become legacy so style unification at this point is only going to apply to code that we are about to deprecate soon. Apologies for the unfortunate timing, we will not be accepting this PR.

PR #481, opened 2024-01-31
Review changes

mcottontensor:
- Tidied up the tsc build process of the common library
- Added WebSocketsTransportNJS implementation to common library for use with node.js applications using this lib.
- Fixed issue with offer/answer handling in frontend lib. This was still working due to the way messages get constructed but the code was still incorrect.
- Adding some comments to SignallingProtocol to explain what events are fired on what emitter
- Moved SignallingProtocol from the WebSockets folder to the Protocol folder since it has nothing to do with websockets now.
- Altered the signaling tester docker to use code in repo rather than published packages.


lukehb (changes requested):
PR itself looks great. As discussed, the failing test should use `npm link` instead of remote dependency, so it is testing latest source.

PR #484, opened 2024-02-13
Remove Unreal version from cirrus log

timbotimbo:
## Relevant components:
- [x] Signalling server
- [ ] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
The master, 5.5 and 5.4 branches still log that Cirrus is for Unreal 5.3
`Running Cirrus - The Pixel Streaming reference implementation signalling server for Unreal Engine 5.3`

## Solution
As this version number isn't very important and it has proven easy to forget on updates, just remove it.
`Running Cirrus - The Pixel Streaming reference implementation signalling server for Unreal Engine`

## Documentation
N / A

N / A

### DIFF
--- a/SignallingWebServer/cirrus.js
+++ b/SignallingWebServer/cirrus.js
@@ -246,7 +246,7 @@ if (config.UseHTTPS) {
 	});
 }
 
-console.logColor(logging.Cyan, `Running Cirrus - The Pixel Streaming reference implementation signalling server for Unreal Engine 5.3.`);
+console.logColor(logging.Cyan, `Running Cirrus - The Pixel Streaming reference implementation signalling server for Unreal Engine.`);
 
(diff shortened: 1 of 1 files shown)


lukehb:
@timbotimbo We had a chat internally, yep, we think removing it is okay. What we doing right now is working on https://github.com/EpicGames/PixelStreamingInfrastructure/issues/10 - so there will be a completely new Cirrus soon anyway so we are not too concerned about removing the version logging from the existing Cirrus. In the new Cirrus we are thinking we will log the package.json version of the signalling server, as this is useful for debugging, but not mention any specific pairing with a version of UE.

CC @mcottontensor As a note to add the proposed logging of version number into the new cirrus.

PR #486, opened 2024-02-14
Fix touch events not working on iOS

### LINKED ISSUE #376: [BUG] - Gesture interaction does not work in UE5.1 with iOS
**UE Version:** 
UE 5.1.1

**Frontend Version:** 
UE5.2

**Problem component**
Frontend, Pixel Streaming C++ plugin

**Description**
Gestures such as pinch and zoom (and even single click) does not work with Enhanced Input and raw input in iOS devices but work in android device. What I noticed while debugging is the blueprint node "Get Touch Input State" will output "Is Pressed" to be false from iOS and be true from android.

 - Test Device: iPhone 13 Pro, iPad Pro 12.9, Samsung Z Flip 5
 - Streaming OS: Windows 10 Enterprise (Version: 21H2)

Belchy06:
## Relevant components:
- [ ] Signalling server
- [X] Frontend library
- [ ] Frontend UI library
- [ ] Matchmaker
- [ ] Platform scripts
- [ ] SFU

## Problem statement:
This PR addresses the problems identified in https://github.com/EpicGames/PixelStreamingInfrastructure/issues/376

## Solution
With the removal of 3D Touch from iPhones since the iPhone XS, the `force` property on the `Touch` event always reports 0. This PR adds a simple check to see if the `force` is 0 for either a touch start or move, and if so it forces it to 1. This is not required on touch ends as a touch can't end with a non-zero force.

### DIFF
--- a/Frontend/library/src/Inputs/TouchController.ts
+++ b/Frontend/library/src/Inputs/TouchController.ts
@@ -178,7 +178,7 @@ export class TouchController implements ITouchController {
                         coord.x,
                         coord.y,
                         this.fingerIds.get(touch.identifier),
-                        this.maxByteValue * touch.force,
+                        this.maxByteValue * (touch.force > 0 ? touch.force : 1),
                         coord.inRange ? 1 : 0
                     ]);
                     break;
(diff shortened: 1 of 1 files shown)


lukehb (approved):
lgtm

